PR #72, opened 2014-01-03
Badges

ricefield:
The badges app provides a full-fledged workflow for the badge team, which produces trainee badges, staff badges, and visitor badges for the FTTA. This module is intended to replace their current workflow which is on top of MS Access. 

- [x] badge upload
- [x] 2x4 badge printout
- [x] individual badge CRUD
- [ ] error handling
- [x] staff badges
- [x] XB badges
- [x] PDF facebooks
- [ ] mobile faceboks

**cleanup:**
- [ ] datatables staticfiles
- [ ] minify staticfiles
- [x] requirements


ap/badges/models.py
```diff
@@ -0,0 +1,42 @@
...
+        ('S', 'Staff'),
+    )
+
+    type = models.CharField(max_length=2, choices=BADGE_TYPES, default='T')
+    original = models.ImageField(upload_to=_image_upload_path)
+    term_created = models.ForeignKey(Term)
+    # thumbnail
+    # badge_size
+
+    # badge information
```
lifeunion: We don't want to add foreign keys for db optimization reasons. This object is not integrally related to users and should not be automatically fetched if I do select_related() on all users. Since this object is only used sparsely by the badge team, should be manually sync'ed with user.

lifeunion:
We are merging!

ricefield:
WOOHOO! way to go @lifeunion and @radicalray !!

Issue #121, opened 2014-08-28
Books Model


allenwebb:
ISBN needs to be a char field of at least length 13 (17 if you want to include the hypens). The reason is ISBN 13 has the last digit for a checksum which can be X for 10.

ricefield:
fixed in #124

Issue #119, opened 2014-08-25
Trainee Vehicles


allenwebb:
The Vehicle Model doesn't cover the following scenarios (that have come up during the data import):
1. There are married couples where two trainees share a vehicle **Fix:** make vehicle a many to many key of trainee
2. There are trainees who may use different vehicles at different times (swap it out with their parent's vehicle, etc) and thus have more than one vehicle **Fix:** make vehicle a many to many key of trainee
3. There isn't a capacity field for the number of seats available (used by the service schedulers and bunk selection) **Fix:** add numeric capacity field
4. There isn't a year field (not necessary but it is present in the old data) **Fix:** add year field
5. Some of the colors imported are longer than the 10 character limit. 20 might be a good size.  **Fix:** increase the length of the color field to 20
6. If a state/province model is added to link the city model to, the state field in this model could point to that model. Also some of the license plates are from Canada so make sure not to limit it to only states in the USA.
7. If the relationship between the trainee and vehicle models is changed to many to many, the license plate field should probably be kept unique for each state.
https://github.com/attendanceproject/djattendance/blob/dev/ap/aputils/models.py#L84

ricefield:
fixed in #124 (minus using `ManyToMany` with `Trainee`)

Issue #118, opened 2014-08-22
Region / State / Province on City needs some kind of validation


allenwebb:
https://github.com/attendanceproject/djattendance/blob/dev/ap/aputils/models.py#L34

We need to validate the region field somehow to reduce the likelihood of garbage being entered such as misspellings etc which would make queries and reports difficult. It might be good to create a model for this. It would allow us to have short codes such as CA for California, etc.

ricefield:
fixed in #124

PR #124, opened 2014-09-26
data migration model fixes 

ricefield:
addresses issues #118 #119 and #121 that @allenwebb brought up while doing data migration.
* fix ISBN on `Books`
* restructuring of state/locality models (there was a redundancy
* updated vehicle model

@ysbecca @ludennis @lifeunion @jonyao if you all could also take a quick look, since these are model changes, I want everyone to make sure these are ok (and that there aren't any major things I'm missing). Thanks!


ricefield:
merging these in. reverted some commits and will clean it up in `dev`

Issue #126, opened 2015-02-02
Admin interface for selecting editing trainees


allenwebb:
On the admin interface there should probably be the option to sort by name and also a way to search or do a "quick lookup" by name.

PR #127, opened 2015-02-10
Fixed performance issue in admin panel for trainee foreignKeys

radicalray:
Issue #128 - Fixed performance issue in admin panel for trainee foreignKeys TA, spouse, and mentor by adding autocomplete box to replace old drop down menu.

Also fixed Issue #126 by adding search box on top of trainee list.


ricefield:
thanks @radicalray! 

btw, @lifeunion i think we need to update our unit tests because i made some changes to the model... sorry =\

PR #136, opened 2015-05-06
adding reflux to the react core.

lifedispenser:
Not ready yet, but a preview of the changes. It is working, so you can test it out. Right now it is in 'dev' mode. The final version needs to have jsxcompile automatic instead of using the native jsxtransformer.


billyeh:
This is in #455

PR #138, opened 2015-05-31
Django 1.8 Migration

radicalray:
Added all the necessary migration files to bridge successfully to django 1.8

### Notes on fresh install on a computer

First time migrating from django 1.6 to 1.7 or above requires you to run 'makemigrations' in every registered app in one go. This command auto maps out all the circular dependencies between models and auto-creates the migration files in each app/migrations folder in the correct sequential order to break up all circular chains. For this migration, I ran:

```
python manage.py makemigrations accounts aputils books classes houses localities rooms services teams terms attendance absent_trainee_roster dailybread leaveslips lifestudies meal_seating schedules syllabus verse_parse --settings=ap.settings.dev
```

At this point, make sure you have a clean blank djattendance db and then run migrate on it (From django 1.7 on, you should be running 'python manage.py migrate' instead of 'python manage.py syncdb'.

```
python manage.py migrate --settings=ap.settings.dev
```

### Notes on install on a computer with the old django server running on it already

To migrate from django 1.6 to django 1.7+ on an existing database with initial data already in it, see following.

Initial migrations on a project can sometimes be troubleshot using --fake-initial

python manage.py migrate --fake-initial
It's new in 1.8. In 1.7, --fake-initial was an implicit default, but explicit in 1.8.

From the Docs:

The --fake-initial option can be used to allow Django to skip an app’s initial migration if all database tables with the names of all models created by all CreateModel operations in that migration already exist. This option is intended for use when first running migrations against a database that preexisted the use of migrations. This option does not, however, check for matching database schema beyond matching table names and so is only safe to use if you are confident that your existing schema matches what is recorded in your initial migration.
https://docs.djangoproject.com/en/1.8/ref/django-admin/#django-admin-option---fake-initial

Running with --fake-inital option makes everything run successfully. 

#### Attention AP developers!!!!
Please do not use the --fake-initial option unless you know what you're doing. Everyone should re-populate their database from scratch to avoid future pains.

### How to add apps+models and make migrations in the future

To add models in the future, please refer to the django 1.8 tutorial/model section. Basically, you no longer need to run makemigrations {appname} explicitly on every new app, simply running makemigrations will scan through all existing migration files and do an auto-diff to generate the correct migration files for your new app and new models.


ap/ap/settings/base.py
```diff
@@ -250,3 +250,9 @@
     'ADMIN_NAME': 'FTTA Admin',
     'LIST_PER_PAGE': 20,
 }
+
+# Settings for graphing SQL Schema
+GRAPH_MODELS = {
+  'all_applications': True,
+  'group_models': True,
+}
```
ricefield: @radicalray what does this do?

ap/accounts/models.py
```diff
@@ -144,7 +145,7 @@ class Profile(models.Model):
     """
 
     # each user should only have one of each profile
-    account = models.OneToOneField(User)
+    account = models.OneToOneField(settings.AUTH_USER_MODEL)
```
ricefield: nice. :+1:

ricefield:
@radicalray maybe update `requirements.txt` to django 1.8? 

@attendanceproject/owners this is a big upgrade, maybe some of you could review this as well? lg2m, let me know if we should merge

ap/ap/settings/base.py
radicalray: This is used to generate the Database Schema picture: https://github.com/attendanceproject/djattendance/wiki/DB-graph!

radicalray:
@ricefield I think requirements already has Django==1.8.1 in this branch.

PR #150, opened 2015-07-06
initial docs

ricefield:
This will be the docs folder where documentation will be written (and hopefully kept up-to-date) in markdown. Mainly this should be module spec and design documentation.
- [ ] project architecture
- [ ] module list w/ desc
- [ ] per-module functional specifications
- [ ] per-module rough design doc


billyeh:
We've moved beyond this pull request since the wiki contains all this information.

We need to consolidate everything we have on dropbox, github, our own wiki, and bitbucket, but that's a whole other issue.

PR #157, opened 2015-11-05
Exams

haileyl:
Current functionality:
* take an exam
* grade an exam (with provisions for both paper and online exams)
* open an exam for retake

There are still a number of to-dos, but the models, at least, are mostly finalized.

Thanks!


radicalray:
:thumbsup Good job Hailey!

ap/exams/exam_helpers.py
```diff
@@ -0,0 +1,105 @@
+from .models import ExamTemplateDescriptor, ExamTemplateSections, ExamResponse
+
+# Returns the section referred to by the args, None if it does not exist
+def get_exam_section(exam_template_pk, section_number):
+	section_pk = "_".join([str(exam_template_pk), str(section_number)])
+
+	try:
+		section = ExamTemplateSections.objects.get(pk=section_pk)
```
ricefield: indentations here look a little big.. are you following PEP8? (4 spaces instead of tabs)

ap/exams/models.py
```diff
@@ -0,0 +1,131 @@
...
+class ExamTemplateDescriptor(models.Model):
+	training_class = models.ForeignKey(Class)
+
+	# an exam is automatically available to be taken by trainees during a given
+	# time frame
+	opens_on = models.DateTimeField(auto_now=False)
+	closes_on = models.DateTimeField(auto_now=False)
+	# TODO: perhaps this should be a length of time instead of hardcoded times?
+
+	is_midterm = models.BooleanField()
```
ricefield: this feels a little clunky. would it be better to simply have an exam type of either midterm or final?

ap/exams/models.py
```diff
@@ -0,0 +1,131 @@
...
+
+	# number of section in the exam
+	section_count = models.IntegerField(default=1)
```
ricefield: @haileyl can you explain the reasoning behind needing the concept of "sections" in the exams model?

ap/exams/models.py
```diff
@@ -0,0 +1,131 @@
...
+
+	# total score is not user set--this is set as questions are added and point
+	# values assigned for each question.
+	total_score = models.IntegerField(default=1)
+
+	def __unicode__(self):
+		return "Exam for %s, [%s]" % (self.training_class, self.training_class.term)
```
ricefield: isn't an instance of this class more accurately an ExamTemplateDescriptor? (rather than an actual instance of an exam)

ap/exams/models.py
```diff
@@ -0,0 +1,131 @@
...
+		general information
+	- ExamTemplateSections: describes a section of an exam, including the
+		questions for the section
+	- Exam: a specific instance of an exam template, holds general information 
+		pertaining to this take of the exam.
+	- ExamResponse: keyed by a combination of exam/question number, holds the 
+		a response to a single exam question
+"""
+
+class ExamTemplateDescriptor(models.Model):
```
ricefield: I personally feel `Descriptor` is unnecessary, and that this class would make more sense as simply `ExamTemplate`

ap/exams/models.py
```diff
@@ -0,0 +1,131 @@
...
+	training_class = models.ForeignKey(Class)
+
+	# an exam is automatically available to be taken by trainees during a given
+	# time frame
+	opens_on = models.DateTimeField(auto_now=False)
+	closes_on = models.DateTimeField(auto_now=False)
+	# TODO: perhaps this should be a length of time instead of hardcoded times?
```
ricefield: You may want to check with the TAs/office sisters on this. My impression is that they would want to be able to manually open/close an exam, rather than having it be determined by time. I believe this is the currently functionality on the attendance server.

ap/exams/models.py
```diff
@@ -0,0 +1,131 @@
...
+
+	def _is_open(self):
+		return time_in_range(self.opens_on, self.closes_on, datetime.datetime.now())
+	is_open = property(_is_open)
+
+class ExamTemplateSections(models.Model):
+	# This field is formed "{pk of ExamTemplateDescriptor}_{section id, 
+	# 1-indexed}".  We will use this to look up rows in this table
+	template_section_key = models.CharField(max_length=100, primary_key=True, 
+											unique=True)
```
ricefield: what happened her?

ap/exams/models.py
```diff
@@ -0,0 +1,131 @@
...
+
+	question_count = models.IntegerField()
+	first_question_index = models.IntegerField(default=1)
+
+	# TODO: this will be replaced by an hstore that will store JSON.
+	# JSON should include at minimum: question, point value
+	# This functionality is available in django 1.8+
+	questions = HStoreField(null=True)
```
ricefield: Not really a fan of this approach, but I suppose having a ton of `Question` objects pointing back to an `ExamTemplate` is not really a good option either. @radicalray any thoughts on this?

ap/lifestudies/urls.py
```diff
@@ -1,5 +1,5 @@
 from django.conf.urls import patterns, url
-from . import views
+from lifestudies import views
```
ricefield: this probably shouldn't be part of this changeset lol.

ap/templates/base.html
```diff
@@ -1,5 +1,4 @@
 {% load staticfiles %}
-{% load bootstrap3 %}
```
ricefield: hmmmm... why are we doing this?

ap/exams/exam_helpers.py
```diff
@@ -0,0 +1,105 @@
...
+		grader_extras += section_grader_extras
+		scores += section_scores
+		current_question += len(responses)
+
+		# TODO(verification): length of responses should be the same as the 
+		# length of grader_extras.  Maybe combine in the section helper fx
+		# so that we can also verify against first_question_index and
+		# question_count under ExamTemplateSections
+
+	return (responses, grader_extras, section_scores)
```
ricefield: Any reason this stuff is not just in `exams/utils.py`? I think it's better if there's just one location for helper/util functions in a given module

ap/exams/models.py
```diff
@@ -0,0 +1,131 @@
...
+DATA MODELS:
+	- ExamTemplateDescriptor: Describes a given exam, including what class and
+		general information
+	- ExamTemplateSections: describes a section of an exam, including the
+		questions for the section
+	- Exam: a specific instance of an exam template, holds general information 
+		pertaining to this take of the exam.
+	- ExamResponse: keyed by a combination of exam/question number, holds the 
+		a response to a single exam question
+"""
```
ricefield: Could you write a more detailed description of your model design somewhere? Can either be here, in the model docstrings, or in the pull request. I just feel like I'm having a hard timing grokking your model design here

ap/exams/models.py
haileyl: The view that we got for the exam module from the office was that eventually all exams should be able to be administered and graded online (including Spirit, TG, etc.), so the breakdown into sections seemed to be reasonable.  It would make doing things like breaking an exam into separate pages (with save exam in progress by section) or having different instructions for different sections or different grade weighting for different sections (read: GOW) natural.

The idea of sections did get put in before the thought of using hstorefields came up, and we didn't revisit the thought afterwards, so I suppose it would be possible to store the necessary information for sections in a questions hstorefield off of the ExamTemplate instead.  I don't really have a preference here.  Is there something that feels "right"-er to you?

ap/exams/models.py
haileyl: Agreed.  This was already in place when I started working on the module.  I'll double-check with the TA/grading sisters to see what they prefer.

There is functionality for makeup/retakes available.  This is what the ExamRetake class tracks.

ap/exams/models.py
haileyl: I've put a document that Becca and I put together last term here:
https://www.dropbox.com/s/nmd3ql5lccw76rz/Exam%20Module%20Design%20Doc.docx?dl=0

It's very incomplete but it does list more information about the requirements/requests that we had from the office, and a bit more information about the models and a tiny bit about why we chose for the models to look the way that they do.

The general idea is:
An ExamTemplate(Descriptor) corresponds one-to-one to the exam.  It keeps track of administrative aspects of the exam, like what class it's related to or whether it's a midterm or final.  The ExamTemplateSection is what keeps track of the actual questions.  Every ExamTemplate(Descriptor) will have at least one ExamTemplateSection, but may have more than one section, depending on the exam.

An Exam is an actual instance of the exam that is related to a specific trainee taking the exam.  If the trainee takes the exam offline, then the score will be manually entered in by a TA or grading sister.  If the trainee takes the exam online, then there will be ExamResponses related to the exam instance.  There should be one ExamResponse per question (though this is something that could be discussed), and the ExamResponse keeps track of what the trainee's response was as well as the grade assigned and any comments that the grader left related to that response.

ExamRetake class is not listed here (oops!) but this keeps track of exams that are opened for retake--it's essentially a relational table that maps an ExamTemplate(Descriptor) to a Trainee, with a little bit of extra data.

ap/exams/models.py
haileyl: Sorry, I'm not quite sure what you mean.  Are you saying to have ExamTemplateMidterm and ExamTemplateFinal?

ricefield: No, I mean to use a choicefield to denote which type of exam this is (in case there is ever an exam that is not a midterm or final, such as MoHL "quizzes"): https://docs.djangoproject.com/en/1.8/ref/models/fields/#field-choices

ricefield:
@haileyl thanks for the comments. I will take a look at the design doc you linked over, maybe later this week. In general, I think this is good. maybe just keep @radicalray in the loop so he can give a secodn opinion on things.

radicalray:
@haileyl Hey Hailey, I checked out your project and ran it. I don't see any UI for TA's to create exams nor for trainees to take it. Am I missing something? This is what I see: https://www.dropbox.com/s/hwwnwz90es5kvkr/Screenshot%202015-11-16%2018.41.19.png?dl=0

ap/exams/models.py
radicalray: More opinionated comments about naming! "ExamTemplate" sounds misleading, like it's a template for TA's to create future new exams. Maybe this should be called "Exam" and the model to keep track of each time a trainee takes an exam should be either attached to their ExamResponse object or be called something like: "ExamAssignment" or "ExamSession" or "ExamTake" or "ExamInstance". Also, I'm pretty sure we can just have a Response model instead of ExamResponse b/c django auto namespace models by app name. (e.g. exams_examreponse) In fact! (@ricefield @haileyl)  I just tested it, you can have same named models as long as they are in different apps. O.O This will make our code a lot more contextualized and concise! So you can change ExamResponse to just Response.

ap/exams/models.py
```diff
@@ -0,0 +1,131 @@
...
+	# 0 indicates first take.
+	retake_number = models.IntegerField(default=0)
+
+	# Calculated and set when grader saves/finalizes exam grading or, if not
+	# taken online, set by the grading sister manually.
+	grade = models.IntegerField(default=0)
+
+class ExamResponse(models.Model):
+	# This field is formed "{pk of Exam}_{pk of trainee}_{question number}"
+	response_key = models.CharField(max_length=100, primary_key=True,
```
radicalray: response_key (also template_section_key) too much compression?, why not just use natural connection (foreignKey?), easier to filter and query in the future when pulling in questions of an exam. Use django object relations exam.section. Same goes for everywhere else this convention is used....

ap/exams/models.py
```diff
@@ -0,0 +1,131 @@
...
+			stats['average'] = float(total)/float(exams.count())
+			stats['minimum'] = minimum
+			stats['maximum'] = maximum
+		return stats
+
+	def _is_open(self):
+		return time_in_range(self.opens_on, self.closes_on, datetime.datetime.now())
+	is_open = property(_is_open)
+
+class ExamTemplateSections(models.Model):
```
radicalray: Section should have optional description (e.g. fill in best choice to question...)

ap/exams/models.py
radicalray: Let's get better description of the hstore field with question types and widgets to handle it. Let's piggyback off of bootstrap widget libraries for the question type definitions. Let's first support: text answers, multiple choice (single / multiple correct answers), true/false, fill in the blank, matching

ap/exams/models.py
```diff
@@ -0,0 +1,131 @@
...
+
+	# TODO: this will be replaced by an hstore that will store JSON
+	# This functionality is available in django 1.8+.
+	response = models.CharField(max_length=10000)
+
+	score = models.IntegerField(blank=True, null=True)
+
+	# TODO: this will be replaced by an hstore that will store JSON.
+	# This functionality is available in django 1.8+.
+	grader_extra = models.CharField(max_length=1000)
```
radicalray: What's grader_extra? Comments on exam? for each question? Need more documentation and spec'ing out. Maybe rename to "grader_metadata". Extra sounds like extra credit

ap/exams/models.py
```diff
@@ -0,0 +1,131 @@
...
+	is_graded = models.BooleanField(default=False)
+
+	# 0 indicates first take.
+	retake_number = models.IntegerField(default=0)
+
+	# Calculated and set when grader saves/finalizes exam grading or, if not
+	# taken online, set by the grading sister manually.
+	grade = models.IntegerField(default=0)
+
+class ExamResponse(models.Model):
```
radicalray: Can we change ExamResponse to Response? In db, seems to be subclassed -> exam_examresponse, I think the namespace is protected. Also, same goes for all the other models that begin with Exam<name> unnecessarily. I doubt AP will get that big: enterprise scale to require super unique class names to prevention namespace collision.

ap/exams/models.py
```diff
@@ -0,0 +1,131 @@
...
+	is_submitted_online = models.BooleanField(default=True)
+	is_complete = models.BooleanField(default=False)
+	is_graded = models.BooleanField(default=False)
+
+	# 0 indicates first take.
+	retake_number = models.IntegerField(default=0)
+
+	# Calculated and set when grader saves/finalizes exam grading or, if not
+	# taken online, set by the grading sister manually.
+	grade = models.IntegerField(default=0)
```
radicalray: grade made need to be doubleField b/c of partial credit. 1.5? Or maybe should be a tuple: (#correct, #total)

ap/exams/models.py
haileyl: Ah hm, that's a thought.  We had considered having a separate scores class to track MoHL/LL or whether or not you'd read for LoS, but I suppose we could potentially use a score field off of exams.  I don't know if I like all of those being classified as an "exam" because in my mind they're still different things, but if you feel strongly in that direction, that would be fine with me.

The purpose of this field, though, is to determine whether this exam score falls into the midterm grade or the final grade.  Our thought was that when the graders/TA sisters were ready to generate the midterm or final grades, they'd be able to pull up all of the scores/exams related to that class for either the midterm or the final and assign weights accordingly, if necessary.  I can add a clearer comment here for that.

ap/exams/models.py
haileyl: Stupid question--is multi-column foreign key supported in django yet?  All resources I pulled up say no, but maybe I'm misunderstanding or missing something.

ap/exams/models.py
ricefield: Ok, that makes sense. In this case, I personally think using a ChoiceField is a clearer naming convention, because semantically, it better conveys what the variable is there for.

ap/exams/models.py
ricefield: No, I'm ok either way. Your explanation makes sense to me. I think you can work that out with @radicalray to see if there's any performance concerns.

ap/exams/models.py
ricefield: ++

ap/exams/static/exams/js/question-display.js
```diff
@@ -0,0 +1,175 @@
...
+    // Question String
+    var question_prompt = document.createElement("input");
+    question_prompt.id = "tb" + index;
+    question_prompt.className = "question form-control";
+    question_prompt.type = "text";
+    question_prompt.value = question.prompt;
+    question_prompt.setAttribute("name", "question-prompt");
+    newDiv.appendChild(question_prompt);
+    var question_point = document.createElement("input");
+    question_point.id = "tb" + index;
```
haileyl: tb+index is used twice for id.  If we want to move toward packaging on the javascript side when we have more question types, maybe consider making this different.  Not super important to change right now though.

ap/exams/static/exams/js/question-display.js
```diff
@@ -0,0 +1,175 @@
...
+    question_point.className = "question form-control";
+    question_point.type = "text";
+    question_point.value = question.points;
+    question_point.setAttribute("name", "question-point");
+    newDiv.appendChild(question_point);
+
+    //Create array of question types
+    var array = ["Essay","Multiple Choice"];
```
haileyl: Don't expose other types until we're ready to use them

radicalray:
Travis is breaking, starting a new pr

PR #158, opened 2015-11-16
initial work on adding django-countries to replace aputils Country model

davidhtien:
replaced `aputils.models.Country` with package `django-countries`: https://github.com/SmileyChris/django-countries

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ venv
 *.sublime-project
 *.sublime-workspace
 ap/media/badges/**
+ENV
+.vagrant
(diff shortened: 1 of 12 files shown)


ricefield:
@radicalray what do you think? I kind of prefer this, as it's cleaner and probably more thorough than what we had in place prior (I think the impetus behind this was to make bootstrapping the project a lot easier—no need for country fixture). do you have any concerns? maybe from the performance side?

.gitignore
```diff
@@ -9,3 +9,5 @@ venv
 *.sublime-project
 *.sublime-workspace
 ap/media/badges/**
+ENV
+.vagrant
```
radicalray: get rid of these 2, since we're not using vagrant anymore

davidhtien:
Hey can we get this merged in?

.gitignore
davidhtien: I don't think its a big deal some people still have vagrant folders on their machines locally doesn't hurt to keep things like this in right?

ricefield:
🙌

PR #162, opened 2015-12-12
Celery

davidhtien:
@radicalray @ricefield @lifedispenser 
Using RabbitMQ as the broker, djcelery to hook into django ORM and postgres, flower as gui monitor
In order to test you must install the celery related requirements in prod.txt as well as RabbitMQ

In one terminal
'sudo rabbitmq-server' or 'sudo rabbitmq-server -detached'

In another
'python ap/manage.py celeryd --verbosity=2 --loglevel=DEBUG'

In yet another
'python ap/manage.py celerybeat --verbosity=2 --loglevel=DEBUG'

If you want to use flower
'flower  beat --address=127.0.0.1 --port=5555'

tasks can be added to any installed app in a tasks.py file with the decorators @task or @periodic_task


davidhtien:
Fixed assign discipline task which currently takes 0.04 or 0.05 sec to run for one trainee with a very simplistic schedule, this seems slow

travis build is failing because we are running pip install for dev but I have placed the dependencies in prod

ap/ap/celery.py
```diff
@@ -0,0 +1,17 @@
+from __future__ import absolute_import
+import os
+from celery import Celery
+
+# Indicate Celery to use the default Django settings module
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'ap.settings.dev')
```
ricefield: what about prod? is it a problem for `settings.dev` to be hardcoded here?

ap/ap/settings/base.py
```diff
@@ -1,8 +1,18 @@
...
 from django.contrib.messages import constants as message_constants
 
+# Settings for djcelery
+CELERY_TASK_SERIALIZER = 'json'
+CELERY_RESULT_SERIALIZER = 'json'
+CELERY_ACCEPT_CONTENT = ['json']
+CELERY_TIMEZONE = 'US/Pacific-New'
+CELERY_RESULT_BACKEND = 'djcelery.backends.database:DatabaseBackend'
+CELERYBEAT_SCHEDULER = 'djcelery.schedulers.DatabaseScheduler'
+
```
ricefield: can we move this further down in the file? or does it need to be up here?

ap/lifestudies/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+#     now = datetime.now()
+#     result = now.day + now.minute
+#     logger.info("Task finished: result = %i" % result)
+
+# currently takes between 0.04 and 0.05 sec to run for one trainee
+# @periodic_task(run_every=(crontab(hour="10", minute="0", day_of_week="6"))) #Run every Lord's Day at 10pm
+@periodic_task(run_every=(crontab(hour="*", minute="*", day_of_week="*"))) #Run every minute
+def assignDiscipline_periodic():
+    logger.info("Start task")
+    term = Term.objects.last()
```
ricefield: not sure how robust this call is (i.e. if people create future terms in advance). `term.models.Term` has a static method `current_term()` that is probably better to use.

ap/lifestudies/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+    period = Period(term).period_of_date(date.today())
+    for trainee in Trainee.objects.all():
```
ricefield: This should be filtered by `active=true` rather than just getting every trainee (which would get every trainee including past trainees who have graduated

ap/ap/celery.py
davidhtien: good point it should probably be prod, I don't know if there's a way to dynamically detect what the environment is?

ap/ap/settings/base.py
davidhtien: I'm pretty sure we can move it down

ap/lifestudies/tasks.py
davidhtien: good point

ap/lifestudies/tasks.py
davidhtien: yeah haha I wrote that super fast totally forgot to go back and change it

PR #199, opened 2016-05-02
Attendance redux

davidhtien:
Wow this is a lot of commits, but attendance is mostly done, still need to do group leave slips and small ui polishing


davidhtien:
@lifedispenser could you take a look?

@radicalray @ricefield

ricefield:
@davidhtien you might want to pull dev and resolve conflicts.

lifedispenser:
Can we remove all the deprecated and unneeded files? They will confuse future devs.

lifedispenser:
Here is a rough draft of the readme that should be in ap/static/js

## How to work in React:
Please see the wiki for using react-gulp to do hot loading dev.
(talk about advantages of hot loading in quick bulelts)
(talk about what you can't do with it combining with django)

## Add js files to the dom through bundle

Why do this? -> auto minifies and concats into one file, saving space.

1) through bower:
(instructions go here)

2) if you need to add manually:
(instructions go here)

lifedispenser:
Let's go through this Saturday when I'm in. We can go through it together and remove more unneeded files. As to the actual code - I'm not sure how to code review since it is a lot of commits - so probably best is to have you walk me through the code that you've written here. 

Thanks DT for your hard work!

lifedispenser:
David and I went through the codebase together. There are some residual scripts that haven't been deleted yet because other pages rely on them (schedule_detail.html, I'm looking at you), and the grunt build needs to be revamped eventually but the code for the most part is straightforward. I'm going to merge it as soon as I get the confirmation from @davidhtien that the previous travis build error has been resolved.

davidhtien:
@lifedispenser should be good to go now

PR #202, opened 2016-05-05
Fixing migrations files

haileyl:
Hit the problem when pulling dev changes that too many of us are trashing migrations files to fix problems, which makes merge almost impossible, without re-trashing files.  Freshly trash migrations files now, so we can all start from a clean place.

After this goes through, every branch will have to:
1. pull the change that deletes all the migrations files (git pull remote e1fee4a)
2. delete any additional migrations files that have been added
3. delete djattendance database (there are ways to get around this, but if the data in your database is semi-insignificant, would recommend just deleting the database)
4. pull the change that adds migrations files again (git pull remote 1841da1)
5. run makemigrations
6. run migrate

There are additional problems around what happens when we continue to make changes on models.py/add migrations files and how those should get merged in as we make changes simultaneously, and I don't have a really good solution for that yet.  Thoughts?

Tagging people that are actively developing/might be interested.  Please add-tag anyone that I've missed.  Thanks!
@radicalray @ricefield @lifedispenser @amarisch @czhaocmu @burricks @davidhtien @leonardlan @schai6 @snowboarder003


radicalray:
LGTM. Ask everyone to pull this Sat.

Issue #207, opened 2016-05-14
Violation of DRY in life studies model


burricks:
While removing logic from the templates in the life studies module, in a couple of small areas, I had some repeated code between two files: 'templates/discipline_list.html' and 'templates/trainee/trainee_discipline_list.html'. Not really sure how to modularize it without repeating myself somewhere else in the code.

burricks:
Fixed

radicalray:
@burricks For future reference, please reference commit + branch / PR issues are fixed so we can track it in the future. Thanks!

PR #209, opened 2016-05-14
Permissions

burricks:
Permissions applied to the life studies model. One issue I may come back to think about if there's time is described [here](https://github.com/attendanceproject/djattendance/issues/207). It involves two cases of repeated code.

TA Permissions operate on the basis that all TA's are superusers


radicalray:
@burricks Hailey is rewriting our trainee models to merge all the sub-tables into one. Will that affect this PR?

burricks:
Yes it will. I'll have to alter these permissions based on the new model.

radicalray:
Ok, cool. I hope to finish it today. Keep you posted.

radicalray:
@burricks I fixed all the migration errors. Can you pull the latest changes and test to make sure all the permissions are working? 

----
Sidenote: Can we give superuser implicit permission to all views and then just add app/specific permissions to view lifestudy modules? i.e. lifestudy_admin_permission or something? I'm worried about roles with so many implicit permissions it's hard to keep track of later since it's distributed across so many view files. What do you think?

PR #211, opened 2016-05-15
permissions added to badges module

burricks:
This permission assumes that all trainees on the 'badges' service will be added to the 'badges' group
@lifedispenser @radicalray


radicalray:
@burricks for this line:   raise_exception = True

Does it throw when exception is wrong? Will it redirect to login page? What's the error response?

Also, how do we handle view methods instead of view classes?

Finally, is there a way to mass-annotate all the badge views instead of having to add the attributes in each one? If we want to lock down whole apps quickly in the future or add a lot of groups to all the app views?

burricks:
The error response is a 403 response (Access Denied). This should only happen if a trainee finds out or happens to type in a URL that only TA's or other trainees have access to. They shouldn't be able to navigate to any forbidden pages just by clicking on the website. I can easily change it, but I found this to be the least confusing.

I'll need to add permissions to the methods. I think I incorrectly assumed that those were taken of by the class above them. 

I've looked into it but have not yet found a way to mass-annotate all of these views without downloading a shady app on github that no one else seems to be using, and that no one has talked about since 2014.

radicalray:
Hey @burricks , I worry a 403 error may let the attacker know that such a page actually exists and that they just need to forge the right permission to get through the security wall (during their vast amount of free time in the training haha). The current behavior on the old server is to log them out and then redirect them to a login page. I think that may be a more a secure wall.

When I mentioned methods, I meant: 

```
def myView(request):
  if request.POST:
     # logic to handle post request
```

Is that what you mean?

For mass-annotation, can you do something like:

```
for v in views.*:
  v.group_required = 'badges'
```

Grab me if this is not making sense. We don't need to have a drawn-out convo on github. thanks!

ruthnan:
I did what @radicalray suggested, but I didn't figure out how to "mass-annotate" the various views. This should be good as far as taking the user to the login page if they don't have badge group permissions, but it seems to think a bit - I don't quite understand his concern about forging permissions. Wouldn't logging them out also let them know that this page exists?

PR #220, opened 2016-05-28
Prod - Deleted all migrations files

radicalray:
Deleted all migrations files so we stop having migration conflict errors. Will track migration files separately.


radicalray:
Merging! All migration files will no longer be tracked.

PR #231, opened 2016-05-30
set bootstrap grid back to 12 columns but still 24 columns for attendance view

davidhtien:
needed to change static and template directories in base.py to get things to work also for some reason

@radicalray


ap/ap/settings/base.py
```diff
@@ -71,7 +71,7 @@
     # Put strings here, like "/home/html/static" or "C:/www/django/static".
     # Always use forward slashes, even on Windows.
     # Don't forget to use absolute paths, not relative paths.
-    os.path.join(SITE_ROOT, 'static'),
+    os.path.join(SITE_ROOT, 'ap/static'),
```
radicalray: @davidhtien What is ap added? It worked before? Also, we should be changing base.py usually. probably dev.py

If you absolutely have to add ap, you should do this:

```
os.path.join(SITE_ROOT, 'ap', 'static')
```

So it concats them with the proper slash depending on os

radicalray:
Only 1 comment. Fix that and merge. looks good! 👍

ap/attendance/templates/attendance/attendance_react.html
```diff
@@ -1,41 +1,186 @@
...
-		// var Trainee = JSON.parse('{{trainee_bb}}'.replace(/&#39;/g, "'").replace(/&quot;/g,'"'));
-	</script>
-	<!-- figure out frontend dependencies in bower -->
-	<script type="text/javascript" src="{% static "js/date_fns.min.js" %}"></script>
-	<script type="text/javascript" src="{% static "js/underscore-min.js" %}"></script>
-	<script type="text/javascript" src="{% static "js/index.js" %}"></script>
-{% endblock %}
-
-<!-- scripts to be added after content is loaded -->
+{% load bootstrap3 %}
```
radicalray: Please pull out html to separate base.html. Do this later.

radicalray:
k, @davidhtien Please pull out HTML to separate base. Thx for the fix. merging!

PR #233, opened 2016-05-31
Import2

haileyl:


radicalray:
Thanks @haileyl ! Really awesome clean code! :) Let's get this merged soon.

ap/accounts/forms.py
```diff
@@ -4,7 +4,7 @@
 class UserForm(forms.ModelForm):
     class Meta:
         model = User
-        fields = ('firstname', 'lastname', 'phone',)
+        fields = ('firstname', 'lastname',)
```
radicalray: I hope this form with phone isn't used anywhere else?

ap/accounts/models.py
```diff
@@ -78,142 +78,169 @@ def create_superuser(self, email, password):
...
+
+    TA = models.ForeignKey('self', related_name='training_assistant', null=True, blank=True)
+    mentor = models.ForeignKey('self', related_name='mentee', null=True,
+                               blank=True)
+
+    locality = models.ManyToManyField(Locality, blank=True)
+
+    team = models.ForeignKey(Team, null=True, blank=True)
+
+    is_hc = models.BooleanField(default=False)
```
radicalray: Maybe we should have another ManyToManyField called role_type? To indicate the permissions this user has.

This way, we can mimick old server permission model:

You create permissions
You create groups
You link users to your groups
You link permissions to your groups

[linked_answer](http://stackoverflow.com/questions/24069572/create-user-roles-and-permissions-extending-the-existing-user-model)

So, we create a model called AccountType to inherit Group from django and then assign user to the HC group.

We would also need to pre-populate the db with permission groups:

[Here](http://stackoverflow.com/questions/25024795/django-1-7-where-to-put-the-code-to-add-groups-programatically)

@burricks , is this correct?

ap/apimport/templates/apimport/term_details.html
```diff
@@ -0,0 +1,139 @@
...
+                {% endwith %}
+                {% endwith %}
+            </div>
+        {% endif %}
+        <button type="submit" class="col-md-offset-4 col-md-2 btn btn-default">Next</button>
+    </fieldset>
+</form>
+
+<script>
+	$(document).ready(function() {
```
radicalray: From looking at @leonardlan 's code in web-access, you can do this:

```
# Needed for JQuery datepicker UI to work
DateInput = partial(forms.DateInput, {'class': 'datepicker'})
class WebAccessRequestCreateForm(forms.ModelForm):
    date_expire = forms.DateField(widget=DateInput())
```

and the js is populated and initialized automatically.

ap/apimport/templates/apimport/term_details.html
```diff
@@ -0,0 +1,139 @@
...
+        var date = new Date('{{start_date}}');
+        $('#startdatepicker').datepicker('setDate', date);
+
+        date.setDate(date.getDate() + (7 * 20) - 1);
+        $('#enddatepicker').datepicker('setDate', date);
+
+        $('#startdatepicker').datepicker()
+            .on("change", function (e) {
+                var date = $(this).val()
+                calculate_dates()
```
radicalray: What does calculate_dates do?

ap/apimport/utils.py
```diff
@@ -0,0 +1,508 @@
...
+
+def date_for_day_of_week(date, day):
+    """ returns the date of the specified day in the week identifid by date.
+        For example, if date 4/9/2016 (Saturday) and 0 (Monday) are passed in,
+        4/4/2016 will be returned. """
+    return datetime.combine(date + timedelta(days=day-date.weekday()), time(0,0))
+
+def term_start_date_from_semiannual(season, year):
+    """ This returns the best-guess term start date for the given semi-annual.
+        Input should follow the form of ("Winter"/"Summer", year) """
```
radicalray: This is great! 👍

ap/apimport/utils.py
```diff
@@ -0,0 +1,508 @@
...
+    term = Term.current_term()
+    if term != None and date.today() < term.end:
+        return True
+
+    return False
+
+def validate_term(start, end, c_init, c_grace, c_periods, c_totalweeks, request):
+    """ Validates the provided information.  Returns true/false based
+        on the validity of the data """
+    success = True
```
radicalray: probably less confusing if called is_success

ap/apimport/utils.py
```diff
@@ -0,0 +1,508 @@
...
+    if mid_term():
+        messages.add_message(request, messages.ERROR,
+            'Cannot start a new term before previous term has ended!')
+        success = False
+
+    return success
+
+def save_file(f, path):
+    """ Saves file with the same filename at the given path relative to the media folder """
+    file_path = settings.MEDIA_ROOT + '\\' + path
```
radicalray: Need to use python to join the paths:

```
UPLOAD_TO = os.path.join(SITE_ROOT, 'media', 'upload')
```

Or else the slashes will not work across different os

ap/apimport/utils.py
```diff
@@ -0,0 +1,508 @@
...
+        address_obj = Address.objects.get(address1=address)
+        return address_obj
+    except Address.DoesNotExist:
+        pass
+
+    addr = city + ", " + state + ", " + country
+    # Key used is related to haileyl's github account
+    args = {'text' : addr,
+            'api_key' : 'search-G5ETZ3Y'}
+    url = 'http://search.mapzen.com/v1/search?' + urlencode(args)
```
radicalray: This is so crazy! haha

ap/apimport/utils.py
```diff
@@ -0,0 +1,508 @@
...
+    user.date_end = term.end
+
+    #TA
+    #Mentor
+
+    try: 
+        # TODO: This needs to be done better, once we get more information about localities
+        locality = Locality.objects.filter(city__name=row['sendingLocality'])[0]
+    except:
+        print "Unable to set locality for trainee: " + row['stName'] + " " + row['lastName']
```
radicalray: Does the front-end actually see this error message? Do we stop execution at this point and stop the import until it's fixed?

ap/apimport/utils.py
```diff
@@ -0,0 +1,508 @@
...
+    return term_minus
+
+def migrate_schedule(schedule):
+    if schedule == None:
+        return
+
+    schedule2 = schedule
+    schedule2.pk = None
+    schedule2.date_created = datetime.now()
+    schedule2.is_locked = False
```
radicalray: What does is_locked mean? Not sure I understand your comments.

ap/apimport/views.py
```diff
@@ -0,0 +1,137 @@
...
+                return self.get(request, *args, **kwargs)
+
+            # Save term to database
+            create_term(season, year, start_date, end_date)
+
+            # Move schedules
+            migrate_schedules()
+
+        # Save out the CSV File
+        request.session['file_path'] = save_file(request.FILES['csvFile'], 'apimport\\csvFiles\\')
```
radicalray: Why all the \\? Will that work on linux filesystems?

ap/apimport/views.py
```diff
@@ -0,0 +1,137 @@
...
+                      request.POST['team_code'], 
+                      request.POST['team_type'], 
+                      request.POST['team_locality'])
+        elif save_type == 'house':
+            save_residence(request.POST['house_name'],
+                           request.POST['house_gender'],
+                           request.POST['house_address'],
+                           request.POST['house_city'],
+                           request.POST['house_zip'])
+    response = {'Todo(apimport2)' : 'Check failure'}
```
radicalray: lolz

ap/aputils/models.py
```diff
@@ -131,23 +131,23 @@ class HomeAddress(Address):
...
-    license_plate = models.CharField(max_length=10)
+    license_plate = models.CharField(max_length=25, blank=True, null=True)
 
-    state = models.CharField(max_length=20)
+    state = models.CharField(max_length=20, blank=True, null=True)
 
     capacity = models.PositiveSmallIntegerField()
 
-    trainee = models.ForeignKey('accounts.Trainee', related_name='vehicles', blank=True, null=True)
+    trainee = models.ForeignKey('accounts.User', related_name='vehicles', blank=True, null=True)
```
radicalray: Should change name to user if model underneath changed so we don't get confused in the future.

ap/schedules/models.py
```diff
@@ -190,12 +188,27 @@ class Schedule(models.Model):
...
+    
+    # If a change comes in part-term, a schedule may differ in different parts of the term.
+    # Parent schedule points to a full-term version of the most-recent schedule, which can be
+    # easily imported to the next term.
+    parent_schedule = models.ForeignKey('self', related_name='parent', null=True, blank=True)
+
+    # is_locked means that for the rest of this term, the trainee set will not change.
+    # If this schedule gets moved to the next term, then we should make sure to reset the
+    # is_locked flag
+    is_locked = models.BooleanField(default=False)
```
radicalray: Is this what you flip on for the new schedule if trainee's schedule changes mid-term?

ap/schedules/models.py
```diff
@@ -190,12 +188,27 @@ class Schedule(models.Model):
...
+
+    term = models.ForeignKey(Term, null=True, blank=True)
+
+    query_filter = models.ForeignKey(QueryFilter, null=True, blank=True)
```
radicalray: Is this query_filter used to indicate what users at attached to this schedule? How does it play with the trainees field?

ap/schedules/models.py
```diff
@@ -242,4 +255,88 @@ def get_absolute_url(self):
...
+        was_active = False
+        for i in range(0, week + 1):
+            if self.active_in_week(i):
+                was_active = True
+                break
+
+        willbe_active = False
+        for i in range(week + 1, 20):
+            if self.active_in_week(i):
+                willbe_active = True
```
radicalray: You forgot to break here.

*cringe*

self.active_in_week runs this every time:

```
weeks = [int(x) for x in self.weeks.split(',')]
```

Can we just do that manually here and make a set?

```
from sets import Set

weeks_set = Set(weeks)

# then you can check to see if there's intersection

len(Set(range(0, week+1)).intersection(weeks_set)) > 0
```

ap/schedules/models.py
```diff
@@ -242,4 +255,88 @@ def get_absolute_url(self):
...
+            sched_weeks = [int(x) for x in self.weeks.split(',')]
+            s1_weeks = []
+            s2_weeks = []
+            for x in sched_weeks:
+                if x <= week:
+                    s1_weeks.append(x)
+                else:
+                    s2_weeks.append(x)
+
+            s1.weeks = "[" + ",".join(s1_weeks) + "]"
```
radicalray: I think ```str(s1_weeks)``` produces the same result and much safer

ap/terms/models.py
```diff
@@ -68,7 +68,11 @@ def current_term():
         except ObjectDoesNotExist:
             logging.critical('Could not find any terms marked as the current term!')
             # try to return term by date (will not work for interim)
-            return Term.objects.get(Q(start__lte=datetime.date.today()), Q(end__gte=datetime.date.today()))
+            try:
+                return Term.objects.get(Q(start__lte=datetime.date.today()), Q(end__gte=datetime.date.today()))
```
radicalray: There's this thing:

You should save datetime.date.today() to a variable and then compare, in the event their evaluation took place in between 2 days. lolz

ap/accounts/forms.py
haileyl: I didn't see UserForm being called anywhere, and there's only one other reference to phone that I found.

ap/apimport/utils.py
haileyl: No we don't.  This is for logging--we actually pre-check locality before we even start the import, so this should never get hit.  One of the remaining work items for import is to add extensive logging at different levels so that whoever is running import can investigate/fix issues if they come up.

ap/accounts/models.py
davidhtien: Do we have to worry about other trainee roles here? Like attendance monitor and ypc monitor?

haileyl: Okay to do this in the next PR?

ap/apimport/templates/apimport/term_details.html
haileyl: @amarisch

ap/apimport/templates/apimport/term_details.html
haileyl: @amarisch where is this?

ap/apimport/templates/apimport/term_details.html
```diff
@@ -0,0 +1,139 @@
...
+
+
+{% endblock %}
+
+{% block title %}Term Details{% endblock %}
+
+{% block content %}
+
+<head>
+<script src="//cdnjs.cloudflare.com/ajax/libs/bootstrap-datepicker/1.3.0/js/bootstrap-datepicker.min.js"></script>
```
davidhtien: I'm a little confused by this syntax for src, is this legit? Just wanted to check

ricefield: yeah, `//` is a relative operator, it uses whatever existing scheme is in place (http://stackoverflow.com/questions/3583103/network-path-reference-uri-scheme-relative-urls)

ap/apimport/utils.py
```diff
@@ -0,0 +1,508 @@
+import csv
```
davidhtien: Wow this is a beast of a file, I didn't comb through it very thoroughly in my first pass, don't have the steam maybe we could get more eyes at least for this one file? @lifedispenser @ricefield

ap/apimport/templates/apimport/term_details.html
radicalray: Oh boy, good catch! We should not be cloudflaring js library files. This is a local network server, haha they won't be able to get those files. >.<

Ray

ap/accounts/models.py
radicalray: Yes, let's make an issue + TODO in the code to be changed later when @burricks pushes in all the permission modules.

ap/apimport/utils.py
```diff
@@ -0,0 +1,508 @@
...
+    residences = []
+    with open(file_path, 'r') as f:
+        reader = csv.DictReader(f)
+        for row in reader:
+            # sendingLocality, teamID, residenceID, officeID
+            if (not check_sending_locality(row['sendingLocality'])) \
+                and (not row['sendingLocality'] in localities):
+
+                if fake:
+                    save_locality(row['sendingLocality'], 1, 'US')
```
radicalray: Is this for testing? fake boolean?

ap/apimport/utils.py
```diff
@@ -0,0 +1,508 @@
...
+    except Address.DoesNotExist:
+        pass
+
+    addr = city + ", " + state + ", " + country
+    # Key used is related to haileyl's github account
+    args = {'text' : addr,
+            'api_key' : 'search-G5ETZ3Y'}
+    url = 'http://search.mapzen.com/v1/search?' + urlencode(args)
+
+    r = requests.get(url)
```
radicalray: If this api breaks in the future (aka, our api_key stops working b/c they change their policies or how likely is it for the company to fold?), is there fallback we can use?

OKAY! I finally found a viable alternative!

http://www.gisgraphy.com/free-access.htm

No api-key required! See this page for how well it performs compared to google: 
http://www.gisgraphy.com/compare/

If we ever wanted to go crazy and install a local address-parsing server: https://github.com/gisgraphy/gisgraphy

Here's an offline address parser we can default to if all the api's broke:
https://github.com/SwoopSearch/pyaddress

ap/apimport/utils.py
```diff
@@ -0,0 +1,508 @@
...
+            user = User(email=row['email'], office_id=row['officeID'])
+
+    user.save()
+
+    if row['residenceID'] == 'COMMUTER':
+        user.type = 'C'
+    else:
+        user.type = 'R'
+
+    user.firstname = row['stName']
```
radicalray: the column name is stName? O.O

Maybe in the future, we can really be fancy and allow user mapping of column to our predefined variables.

ap/accounts/models.py
haileyl: @davidhtien, the only one we read in at this point is HC.  AM and YPC monitors are set separately.

ap/apimport/utils.py
haileyl: Yes for testing.  I'll delete it once we have the server up and running.  It's just a quick way to get past manually creating all of the localities, all of the teams, and all of the houses.

ap/apimport/utils.py
haileyl: Took a look at gisgraphy and it looks reasonable in all the things I ran into problems relating to previously.  I think mapzen is pretty safe though as well.  How concerned are we, really?  Mapzen is based on open source tools, so I'm fairly confident that it will be free to use at least at the capacity that we are running for a while.  The key is available through github, so even if my key expires, it should be trivial to grab a new one?

radicalray: How often do their key expire? Maybe we should catch such an error and let the dev know the api broke and needs updating. I would prefer this feature to work consistently with minimal tweaking/fixing by devs in the future b/c they'll most likely be under a lot of pressure if they need to fix importer on a 1 week deadline during pre-training week.

I read their policy:

Mapzen Search allows you a maximum of:

6 requests per second
30,000 requests per day

Is the 6 req/sec cap a concern for us? Could we accidentally request too fast programmatically? There's no checks and our production server is really fast...haha

They also mentioned that the rates cap are liable to change at any time in the future without notification.

haileyl: It hasn't been a problem with the import files that I've been using, and it should be less of a problem as time goes on.  I don't think our numbers will ever exceed their rates cap though.  Even if we have to run importer two or three times because of some mistake that we hit, we shouldn't get above 1500 requests per day, and I think mapzen caches hits as well, so we shouldn't ever get that high.

I can take a look at gisgraphy later this week/next week and see if it's something that we can swap in.  At first glance it looks like it'll play nicely with what we have, but I'll have to try it on a couple of different addresses before I'll be sure.

ap/apimport/utils.py
haileyl: It's mostly used in the mid-term split-schedule case (e.g. a user changes team mid-term).  If we do a mid-term import to update HCs or some other variable, we may re-run schedule assignment, but for an is_locked schedule, we wouldn't want to run schedule assignment.  It's likely that we can figure this out through date checking or some other heuristic, but I like having the variable to be safe.

ap/schedules/models.py
haileyl: flip on for the old copy of the schedule.  e.g. if the schedule split into schedules A and B, where A covers weeks 1-10 and B 11-end, A would be locked, but B would not be.  I can make the comment a bit clearer.

ap/schedules/models.py
haileyl: query_filter is used to assign trainees to schedules in assign_trainees_to_schedule. The interesting part (the part that has actually been tested so far) is the last two lines of that function.  query_filter and the trainees actually assigned to the schedule should always be in sync unless the schedule has is_locked set to true.

ap/terms/models.py
haileyl: if this ever happens...

ap/apimport/templates/apimport/term_details.html
haileyl: @amarisch

ap/apimport/templates/apimport/term_details.html
haileyl: @amarisch

radicalray:
@amarisch Is everything fixed? If so, tag me and let me know.

radicalray:
@haileyl Ok, I finally got to test the full module! Here are some notes/questions:

 - State field in form is a dropdown and is empty if state table in db is empty. Maybe states should be hardcoded and should only appear if they choose USA?
     - USA should be top of dropdown list for countries for ease?
 - In the tabs for importing Locality/Team/Houses, there isn't a tab for showing the new trainees being imported? Pretty scary if you don't know who got imported or not.
 - refresh button. What' the intention of the refresh button on under each tab?
 - When I saved everything, there was no congratulations screen. In fact, when I checked the db, no users were imported. Is that a mistake? I couldn't find any methods in the code that created new users either. :(

haileyl:
@radicalray see responses to your comments inline.

radicalray:
Thanks, @haileyl ! Merging. Fix issues later.

PR #267, opened 2016-06-17
Permission groups

burricks:
This code programmatically creates permission groups. The code is run after every call to 'migrate', whether or not there are any new migrations or if the programmer is working on a fresh database or not. This provides a way for dev's to standardize the creation of all the permission groups to prevent complications for developers in the future - so that they won't have to recreate all the permission groups from a database while writing or testing code. 

Of course, this is an initial proposition for the permission groups we need. They are flexible and can be altered. For an explanation of how these permission groups are to be used in code, see the wiki page named 'Permissions'.


ap/accounts/admin.py
```diff
@@ -340,3 +340,6 @@ def save_model(self, request, obj, form, change):
 admin.site.register(User, APUserAdmin)
 admin.site.register(Trainee, TraineeAdmin)
 admin.site.register(TrainingAssistant, TrainingAssistantAdmin)
+
+# Unregister Group from Admin site, to restrict Group alterations to dev's
+admin.site.unregister(Group)
```
burricks: The 'Group' model is unregistered from the Admin site so that changes can be restricted to dev's in the code. This prevents the groups from existing solely in the database.

PR #268, opened 2016-06-17
Fix seating chart

radicalray:
Rewrote jquery-seating-charts to handle focus/blur events correctly now.

Supports keypress navigation, also bootstrap popover.

Also, absorbed grid object into library initialization for seating chart creation. No can handle arbitrary amount of data stored in each cell.

TODO: Doesn't have any too utility methods to give native access to grid object within seating chart yet.

Now, we locally maintain our fork of seating chart: https://github.com/attendanceproject/jQuery-Seat-Charts


radicalray:
Merging! @jontey @davidhtien

PR #282, opened 2016-06-22
Fixed ajax bug for update books page, fixed css, cleaned up js code a…

radicalray:
… little bit

Also consolidated status flashing behavior to one function.


radicalray:
Merging!

PR #298, opened 2016-06-30
Added ability to backup scripts as well

radicalray:


radicalray:
Good job @snowboarder003 !!! Come on!

PR #328, opened 2016-07-19
Fixing Admin-related things

radicalray:
Goal of this PR:
 - [x] Swap out states object to be pythonic rather than db dependent
 - [x] added good defaults to country
 - [x] changed import to best of my abilities. @haileyl please check if I changed import correctly to accommodate changing state.

 - [x] Admin page for adding people to house/permission groups
 - [x] Fixed Roll js bug (ever expanding width from click)
 - [x] pulled out trainee_utils to prevent cyclical dependencies.


radicalray:
Tested apimport works. Merging. @haileyl Don't forget to pull origin dev before you merge/continue to develop.

PR #359, opened 2016-07-26
Roll Bugs (1st PR)

jontey:
Things fixed in this PR
- [x] Fix week selector for rolls input
- [x] Fix 1st and 2nd year filter for roll table views
- [x] Only show 1st year trainees and 2nd year that are audited
- [x] Fix click events not registering on PCs with touchscreens
- [x] Smart redraw function for Class Roll

Smaller PR's to make code review a little easier @radicalray


ap/attendance/static/attendance/js/seating-controller.js
```diff
@@ -514,10 +499,59 @@ var SeatController = {
       if (body.height() < sm.height()) {
         body.height(sm.height());
       }
+  },
 
+  // Smarter draw function.. Instead of redrawing everything
+  update: function (trainee){
+  	var t = SeatController;
+  	var i = trainee.y;
```
radicalray: What if a trainee is assigned to multiple seats by accident? What will be the result behavior?

ap/attendance/static/attendance/js/seating-controller.js
```diff
@@ -514,10 +499,59 @@ var SeatController = {
...
+    if(trainee.status == 'U'){
+      t.show_notes(seat, j, i);
+    }
+  },
 
-
+  draw_node: function(node, seat){
+  	var t = SeatController;
+  	if(node && seat){
+      if(seat.gender == t.gender){
```
radicalray: We can indicate a seat is a certain gender? O.O Didn't know we had that feature.

ap/attendance/static/attendance/js/seating-controller.js
```diff
@@ -514,10 +499,59 @@ var SeatController = {
...
+        node.html("<b>"+seat.name+"</b>");
+        node.attr('title', seat.notes);
+        if(seat.attending){
+        	node.removeClass('roll-absent uniform_tardies uniform roll-tardy left-class');
```
radicalray: Not sure what this code does? ^

ap/attendance/templates/attendance/roll_table.html
```diff
@@ -193,8 +193,8 @@
     }
     $(function (){
       setWeeks();
-          $("#select_menu option[id='week_{{ current_week }}']").attr("selected", "selected");
-          setDatesforWeek({{ current_week }});
+      $("#select_menu option[id='week_{{ current_week }}']").attr("selected", "selected");
```
radicalray: In general, try not to print django variables directly into js code. Pull to the top for easier configuration by other devs (backend) in the future.

ap/attendance/views.py
```diff
@@ -233,7 +234,7 @@ class HouseRollsView(TableRollsView):
     def get_context_data(self, **kwargs):
         user = self.request.user
         trainee = trainee_from_user(user)
-        kwargs['trainees'] = Trainee.objects.filter(house=trainee.house)
+        kwargs['trainees'] = Trainee.objects.filter(house=trainee.house).filter(Q(self_attendance=False,current_term__gt=2)|Q(current_term__lte=2))
```
radicalray: Good catch! 👍

ap/static/js/jquery.seat-charts.js
```diff
@@ -170,12 +170,23 @@
 
     //true -> deep copy!
     $.extend(true, settings, setup);
-
-    var events = Modernizr.touchevents ? 'touchstart touchend' : 'mouseup mousedown contextmenu';
+    var events = 'touchstart touchend mouseup mousedown contextmenu';
```
radicalray: Catching mouseup/mousedown on a touch device is super buggy with different behaviors across different phones. Are you sure this will be a safe fix? Previous fix was tested across 6 different phones.

ap/terms/models.py
```diff
@@ -111,7 +111,7 @@ def is_date_within_term(self, date):
         return date >= self.start and date <= self.end
 
     def startdate_of_week(self, week):
-        return self.start + timedelta(weeks=(week-1))
+        return self.start + timedelta(weeks=week)
```
radicalray: Great catch! This was what was causing the meal roll blank week0 bug!

Maybe we should update enddate_of_week too and add in comments clarifying what week ranges it's expecting as input args. Are the period calculation correct? Check the whole file.

ap/static/js/jquery.seat-charts.js
jontey: That is why I have the preventdefault. Then the mouse up/down event won't
trigger if the user uses the touch event.

ap/attendance/static/attendance/js/seating-controller.js
jontey: Basically fills in the trainee name and add proper css classes.
I do the check for valid nodes and trainee here because the same logic is run everytime you want to redraw a particular node.

ap/attendance/static/attendance/js/seating-controller.js
jontey: The result wont be consistent if you have a trainee assigned multiple seat... but maybe implement that check when creating seating chart?

That brings up another problem... If the ajax call returns and we try to update the value but the user changed the section so the trainee's seat either moves or is not shown anymore. We could be updating with stale info.... Need to implement check for this case

radicalray: Good edge case you bring up. Yea, you'll need to write a translation function that translates x, y to currently displayed based on which section open. Maybe you can you reuse some logic you wrote before to calculate section display?

Also, When sections have overlapping seats, the toggle gets kind of buggy, can't really pin it down to what's the cause.

@rayoowoo is not testing your PR running on my computer and may have further comments.

rayoowoo:
- [ ] viewing the page - selecting any week will cause the previous week to be displayed. (e.g. if you select week 8 in the dropdown menu, the information for week 7 is displayed. and your selection will display week 7 as well.) 
- [ ] inputting roll statuses - finalize doesn't work
- [ ] meal rolls - roll statuses are not being saved
- [ ] not all the trainees in the seating chart appear on the class/study rolls

jontey:
Added leaveslip display for roll entry module

PR #363, opened 2016-07-27
Exams+Class Bugfix

radicalray:
Consolidated class models, rewrote much of exam frontend, performance tweak as well

 - [x] Performance tweaking for grades page (GET + POST)
 - [x] Rewrote exams frontend (HTML + JS + CSS)
 - [x] Fixed some conversion bugs (convert string to boolean, unquoted HTML attributes ```(values={{ blah }}))```


ap/exams/utils.py
```diff
@@ -39,23 +41,23 @@ def get_exam_questions(exam):
...
+        # per section are vaguely correct whenever we have an exam that has
         # when we start having exams with more than one section.
     return questions
 
-def get_exam_questions_for_section(exam, section_id):
-    section = get_exam_section(exam, section_id)
-    questions = []
-    if (section == None):
-        return None
+# def get_exam_questions_for_section(exam, section_id):
```
radicalray: Delete duplicate entry.

PR #368, opened 2016-07-29
Ss Merge models to get rid of conflicts

radicalray:


radicalray:
Merging in the models!

PR #374, opened 2016-08-09
Admin

burricks:
These few lines of code are a simple way to bring the menu from admin that would display for the given user into the template in the front-end.


billyeh:
We basically just use panels and other links to link to admin now I think

PR #388, opened 2016-10-15
Added required version for faker to install

jontey:
Removed migration files from repo (It is now tracked on another repo)


radicalray:
We no longer use faker, feel free to throw it out so we don't waste more time in the future trying to support it.

jontey:
OK. Do you remember what David Tien use to generate the default data?

radicalray:
Use this: https://github.com/attendanceproject/djattendance/wiki/Generating-Dummy-Data

Issue #261, opened 2016-06-11
Life study alerts


allisonfaith:
multiple life study alerts for the same life study assignment on home page

radicalray:
@allisonfaith  Screenshots and more description please.

allisonfaith:
![life study notification bug](https://cloud.githubusercontent.com/assets/19786868/15987701/2e23b284-2fe7-11e6-8efd-0026fb35bb47.PNG)

Even when I dismiss the alerts and refresh the home page, there are still multiple alerts.

jontey:
I thought that was a bug on my database. I can take a look at it

rayoowoo:
was this resolved? @jontey @allisonfaith

PR #391, opened 2016-11-11
fix messages appearing twice

billyeh:
`bootstrap_messages` renders each message already so for each message in there it was creating a whole list of messages again. This fixes it.

### DIFF
--- a/ap/templates/index.html
+++ b/ap/templates/index.html
@@ -7,9 +7,7 @@
 
 {% block messages %}
     <div id="messages">
-        {% for message in messages %}
-            {% bootstrap_messages message %}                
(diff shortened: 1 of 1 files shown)


billyeh:
Yeah actually according to this page http://django-bootstrap3.readthedocs.io/en/latest/templatetags.html#bootstrap-messages we don't need to do anything other than include the tag to show all the messages. We wouldn't lose any messages that would have been shown before, for sure.

Thanks! 😀

radicalray:
Awesome! :+1

jontey:
Good catch

jontey:
Fixes #261

Issue #154, opened 2015-09-23
Django admin many-to-many select2 fields buggy


radicalray:
CSS file (select2.css) doesn't load after first page refresh also javascript onchange doesn't fire correctly when adding new option, due to conflicting django versions (1.8 vs. 1.11), currently fixed by manually overriding css in base_admin.html page. Let's find a more elegant solution in the future.

billyeh:
This is fixed by the bower reorg.

PR #392, opened 2016-11-12
use fix from admin page for select on main site

billyeh:
I found this fix on this page already https://github.com/attendanceproject/djattendance/blob/dev/ap/templates/admin/base_site.html but that's for the admin site.

Without this, the select2 box for e.g. on the Create Event page for trainees doesn't work after a submission or refresh. Seems like the scripts have to be re-included to re-initialize the select2 box.


billyeh:
Actually, I wonder we should wait for @lifedispenser's react reorg

radicalray:
His react reorg should not affect admin too much.

billyeh:
Hmm I actually wanted to move these lines of code into the non-admin site from the admin site, so I was thinking it'd be affected by his reorg.

billyeh:
So yeah this is already documented on #154

ap/templates/base.html
```diff
@@ -20,6 +20,9 @@
       {% load bootstrap3 %}
 
       <link href="{% static "css/base.css" %}" rel="stylesheet">
+      {# Quick fix b/x select2 css refuses to load after 2nd page refresh #}
+      <link rel="stylesheet" type="text/css" href="{% static 'django_select2/css/select2.css' %}" media="all">
```
billyeh: I want to add these scripts to the announcements creation and update pages that I'm working on. You're right, that makes a lot more sense. Thanks!

PR #398, opened 2016-11-20
Ss interim beta

radicalray:
Added core service scheduling components:

 - Graph flow algorithim
 - Workers capacity view
 - Service workers capacity/health tweaking
 - Graph-debugging
 - Admin pages
    - Services
    - Workers
    - Exceptions

Left to do:
 - Calendar view for services
 - Designated services input schedules + precepts (rotating schedules) + exceptions
 - Lots of small edge-case functionalities to be covered

@jontey: We may want to merge this in soon if we anticipate the codebase changing to avoid massive PR's/conflicts in the future.


radicalray:
@jontey Please take a look at this pr and let's get this merged in soon to avoid conflicts blowing up in the future.

radicalray:
@jontey are the new commits for deploying to heroku?

radicalray:
Also, we should prob gitignore data.js lest we accidentally check in user data

radicalray:
@jontey I forgot to tell you but I think I wrote ```python manage.py populate_trainees``` type auto-gen functions to auto-gen services. Hope that helps you test locally. That only generates the bulk of weekly-recurring services but I haven't really covered more hidden services that are one-offs (office sisters, etc...) Feel free to add to it.

jontey:
Yes I saw that. I've added all the current trainees on it so it gives a
better feel of the algorithm

PR #423, opened 2016-12-17
Announcements

billyeh:


ap/announcements/templates/announcement_detail/table.html
```diff
@@ -0,0 +1,28 @@
+<table class="table">
+    <tr><td>Status:</td>
+        <td>{{ announcement.get_status_display }}</td></tr>
```
radicalray: Are these tabs or 4 spaces? @billyeh, Can you change all the spacing to 2-space tabs for consistency? Sublime can do that automatically for you (convert space to tabs, change tab size and then convert back to spaces) There might be faster ways to do this in bulk but this is the way I know on top of my head.

ap/announcements/views.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def get_context_data(self, **kwargs):
+        context = super(AnnouncementList, self).get_context_data(**kwargs)
+        context['date'] = self.date
+        context['form'] = AnnouncementDayForm()
+        return context
+
+    def get_queryset(self):
+        announcements = Announcement.objects \
+        .filter(Q(type='CLASS',
```
radicalray: I don't think you need Q, you can just do ```.filter(type='CLASS', status='A', ...)```

ap/aputils/trainee_utils.py
```diff
@@ -1,13 +1,19 @@
 from accounts.models import Trainee
 
 def is_trainee(user):
-    t = user.type
+    try:
```
radicalray: All users should have a type, how can this throw? If user doesn't have a type, then is_trainee is being passed the wrong type of object. This can have undesired behavior. If the template does ```{% if not is_TA(user) %} <-- display trainee info (e.g. user.type) --> {% end %}``` this will cause silent errors.

ap/aputils/trainee_utils.py
```diff
@@ -1,13 +1,19 @@
...
+        t = user.type
+    except:
+        return False
     if t == 'R' or t == 'C' or t == 'S':
         return True
     return False
 
 def is_TA(user):
-    t = user.type
+    try:
```
radicalray: Same goes here.

ap/templates/popups/in_class_popup.html
```diff
@@ -0,0 +1,30 @@
...
+    </script>
+</head>
+
+<body>
+    <div id="in_class_popup" class="modal fade" tabindex="-1" role="dialog" aria-labelledby="in_class_popup" aria-hidden="true">
+        <div class="modal-dialog">
+            <div class="modal-content">
+                <div class="modal-header">
+                    <button type="button" class="close" data-dismiss="modal" aria-hidden="true">&times;</button>
+                    <h4 class="modal-title">You are using the server in class</h4>
```
radicalray: We can make this more personal: ```{% trainee.fullname %}, you are using the server in class```

ap/templates/requests/request_list.html
```diff
@@ -0,0 +1,85 @@
...
+            </div>
+        </div>
+    </div>
+</div>
+
+<h1>{{ item_name|capfirst }} requests</h1>
+
+<a href="{{ create_url }}">Create new {{ item_name }} request</a>
+
+{% regroup object_list by get_status_display as sorted %}
```
radicalray: ^ this is handy

ap/announcements/templates/announcement_list/title.html
```diff
@@ -0,0 +1 @@
+{{ request.get_type_display }}
```
radicalray: Any reason this is a one line file? May be clearer to just have this wherever it's referenced unless you foresee the logic here getting really messy in the future?

ap/announcements/models.py
```diff
@@ -0,0 +1,55 @@
...
+from accounts.models import Trainee
+
+class Announcement(models.Model):
+
+    class Meta:
+        verbose_name = "announcement"
+        ordering = ['-announcement_date']
+
+    ANNOUNCE_STATUS = (
+        ('A', 'Approved'),
```
radicalray: I think you reference this dictionary elsewhere to translate status to text. You can do ```.get_status_display``` to show the display name and then lowercase it for your text. For future references, you can always do: ```. get_ATTR_display``` (See: https://docs.djangoproject.com/en/1.8/ref/models/instances/#django.db.models.Model.get_FOO_display)

ap/announcements/notifications.py
```diff
@@ -0,0 +1,86 @@
...
+from web_access.models import WebRequest
+from terms.models import Term
+from aputils.trainee_utils import is_trainee, trainee_from_user
+
+def get_popups(request):
+    if not is_trainee(request.user):
+        return []
+    trainee = trainee_from_user(request.user)
+    today = datetime.date.today()
+    announcements = Announcement.objects \
```
radicalray: You should try to sort queries here instead of the template. it's way faster (postgres > python)

ap/announcements/notifications.py
```diff
@@ -0,0 +1,86 @@
...
+            is_popup=True) \
+        & (Q(num_trainees=0) | Q(trainees=trainee))) \
+        .exclude(trainees_read=trainee)
+    return list(announcements)
+
+def get_announcements(request):
+    notifications = []
+    if is_trainee(request.user):
+        trainee = trainee_from_user(request.user)
+        notifications.extend(discipline_announcements(trainee))
```
radicalray: What's preventing you from just concat'ing the lists here? ```notifications = discipline_announcments(trainee) + server_announcements(trainee) + ...``` Reads clearer.

ap/announcements/notifications.py
```diff
@@ -0,0 +1,86 @@
...
+        notifications.extend(server_announcements(trainee))
+        notifications.extend(bible_reading_announcements(trainee))
+        notifications.extend(request_statuses(trainee))
+    # sort on severity level of message
+    return sorted(notifications, lambda a, b: b[0] - a[0])
+
+def request_statuses(trainee):
+    requests = []
+    requests.extend(IndividualSlip.objects.filter(trainee=trainee, status='F'))
```
radicalray: same here

ap/announcements/notifications.py
```diff
@@ -0,0 +1,86 @@
...
+        if stats['number_filled'] < 7:
+            unreadWeeks.append(str(w))
+    if not unreadWeeks:
+        return []
+    return [(messages.WARNING, fmtString.format(url=url, week=', '.join(unreadWeeks)))]
+
+def server_announcements(trainee):
+    today = datetime.date.today()
+    announcements = Announcement.objects \
+        .annotate(num_trainees=Count('trainees')) \
```
radicalray: You may want to pull this complex query out and isolate it if things change in the future, it'll be easier to debug and update.

ap/announcements/models.py
```diff
@@ -0,0 +1,55 @@
...
+    ANNOUNCE_TYPE = (
+        ('CLASS', 'In-class'),
+        ('SERVE', 'On the server')
+    )
+
+    status = models.CharField(max_length=1, choices=ANNOUNCE_STATUS, default='P')
+    type = models.CharField(max_length=5, choices=ANNOUNCE_TYPE, default='CLASS')
+
+    date_requested = models.DateField(auto_now_add=True)
+    trainee = models.ForeignKey(Trainee, null=True)
```
radicalray: Is trainee the author of Announcement?, the naming is a little confusing. trainee and then trainees below. Consider rename for clarity?

ap/templates/requests/request_list.html
billyeh: Yeah, hoping we can re-use this for web requests, A/V, room reservation, etc.

ap/announcements/models.py
billyeh: Cool, thanks! Hmm I don't see anywhere where I reference ANNOUNCE_STATUS with a quick grep, though. I'll try and see if I remember anything. Seems pretty inconvenient if I have used this dictionary somewhere.

ap/aputils/trainee_utils.py
billyeh: AnonymousUser which you get if you're not logged in doesn't have a type. I removed the try/catch and just handle this the same way as `exams/context_processor.py` used to, just check if there's a type attr on user. I'm not sure if it's better, maybe what we should do is have a more explicit exception to catch so that we don't have the silent errors.

ap/announcements/templates/announcement_list/title.html
billyeh: This is the alternative view to `announcements/templates/ta_title.html` that gets inserted into a request list. I'm not sure of a better way to do this. It's kind of unfortunate because each "request" model may have a different name for the "type" field which we want to display here. E.g. right now it's `type` for announcements and `reason` for web access. I don't think this'll complicate things too much though.

ap/announcements/templates/announcement_detail/table.html
billyeh: K, changed all the files in this pull request to 2 spaces :) was 4 before in some parts

ap/announcements/notifications.py
billyeh: OK. Sorry, I'm not sure where sort is used in templates that you're referring to. 

The only instance I remember is the regroup in the request list template on request status, which is a little different.

billyeh:
Thank you for your feedback! I did my best to address your comments.

ap/announcements/views.py
```diff
@@ -0,0 +1,162 @@
...
+  model = Announcement
+  template_name = 'announcement_list.html'
+
+  def get_context_data(self, **kwargs):
+    context = super(AnnouncementRequestList, self).get_context_data(**kwargs)
+    context['item_name'] = Announcement._meta.verbose_name
+    context['create_url'] = Announcement.get_create_url()
+    context['detail_template'] = 'announcement_list/description.html'
+    if is_TA(self.request.user):
+      context['item_title_template'] = 'announcement_list/ta_title.html'
```
radicalray: Is the branching logic pulled into the view for security concerns? Not exposing TA view logic in the template? If this is pushed to template as a ```{% if is_TA %} <ta_title.html> {% else %} <title.html> {% end %}```, it'll get rid of the one-line extra template files?

ap/announcements/templates/announcement_list/title.html
radicalray: See comment below

ap/announcements/notifications.py
```diff
@@ -0,0 +1,78 @@
...
+      is_popup=True) \
+    & (Q(num_trainees=0) | Q(trainees_show=trainee))) \
+    .exclude(trainees_read=trainee)
+  return list(announcements)
+
+def get_announcements(request):
+  notifications = []
+  if is_trainee(request.user):
+    trainee = trainee_from_user(request.user)
+    notifications += discipline_announcements(trainee)
```
radicalray: Minor preference: 
```
notifications += discipline_announcements(trainee) + \
                 server_announcements(trainee) + \
                 ...
```                          
May be clearer. I hope the backslash works.

ap/announcements/notifications.py
```diff
@@ -12,75 +12,67 @@
...
+  if not is_trainee(request.user):
+    return []
+  trainee = trainee_from_user(request.user)
+  today = datetime.date.today()
+  announcements = Announcement.objects \
+    .annotate(num_trainees=Count('trainees_show')) \
+    .filter(Q(type='SERVE',
+      status='A',
+      announcement_date__lte=today,
+      is_popup=True) \
```
radicalray: You can make is_popup a boolean argument into ```.announcements_for_today``` to pull this logic out too.

ap/announcements/views.py
```diff
@@ -16,148 +15,148 @@
...
-
-    return redirect('announcements:announcement-request-list')
+  announcement = get_object_or_404(Announcement, pk=id)
+  announcement.status = status
+  announcement.save()
+  announcement = get_object_or_404(Announcement, pk=id)
+  name = announcement.trainee
+  message = "%s's %s web request was " % (name, announcement.get_type_display())
+  if status == 'A':
+    message += 'approved.'
```
radicalray: This could use the ```.get_status_display()``` I mentioned earlier. ```announcement.get_status_display().lowercase()```

ap/aputils/trainee_utils.py
```diff
@@ -1,25 +1,19 @@
...
 def is_TA(user):
-    try:
-        t = user.type
-    except:
-        return False
-    if t == 'T':
-        return True
-    return False
+  t = user.type
+  if t == 'T':
```
radicalray: You can just do ```return t == 'T'```

ap/aputils/trainee_utils.py
```diff
@@ -1,25 +1,19 @@
...
 def is_trainee(user):
-    try:
-        t = user.type
-    except:
-        return False
-    if t == 'R' or t == 'C' or t == 'S':
-        return True
-    return False
+  t = user.type
+  if t in ('R', 'C', 'S'):
```
radicalray: Same here like comment below. Much better. 👍

ap/templates/popups/in_class_popup.html
```diff
@@ -1,30 +1,30 @@
...
-                <div class="modal-footer">
-                    <button type="button" class="btn btn-default" data-dismiss="modal">OK</button>
-                </div>
-            </div>
+  <div id="in_class_popup" class="modal fade" tabindex="-1" role="dialog" aria-labelledby="in_class_popup" aria-hidden="true">
+    <div class="modal-dialog">
+      <div class="modal-content">
+        <div class="modal-header">
+          <button type="button" class="close" data-dismiss="modal" aria-hidden="true">&times;</button>
+          <h4 class="modal-title">{{ trainee.fullname }}, you are using the server in class</h4>
```
radicalray: 👍

ap/announcements/models.py
radicalray: Added comment to point out where you can use this. I think there was another place but don't remember.

ap/announcements/models.py
radicalray: What do you think about renaming trainee to author or creator for clarity? Yea, Leaveslips and web access requests should probably likewise have had better naming as well but we didn't catch it earlier. ;)

ap/announcements/notifications.py
radicalray: I think that's the one ^ I was thinking of but that may not be the one this is referring to. My bad. This is more of a FYI comment then.

ap/announcements/views.py
billyeh: Yeah that's really good, actually, we can get rid of all those if's. Actually, this code was copied from `web_access/views.py`. I think a `Request` base class for web access, announcements, and all the other requests we plan to support would be a good thing.

ap/announcements/views.py
billyeh: If we have a `Request` base class then this will go right in the template. Otherwise, the current one line in the template works for announcements but not web access because it needs `get_reason_display`, not `get_type_display`. (See below suggestion).

ap/announcements/notifications.py
billyeh: Only thing is I didn't plan on using `announcement_end_date` in popup announcements, but I remove that field for popups anyways. So this is great.

ap/announcements/models.py
billyeh: Hmm, I made some comments below. Maybe what we need is an interface for requests. Getters and setters for this kind of thing, rather than relying on the field name being the same, which is what we kind of do now to avoid code duplication.

billyeh:
OK. I think I addressed all your comments this time. Thank you for putting in work on a weekend! :) I went with all the suggestions you had. Maybe when we get around to implementing all these kind of requests we can see if a superclass is a good way to reduce duplicated code, since the workflow is really similar for a lot of these things.

billyeh:
@jontey I think we're ready :)

ap/announcements/notifications.py
```diff
@@ -0,0 +1,70 @@
...
+  trainee = trainee_from_user(request.user)
+  announcements = Announcement.announcements_for_today(trainee, is_popup=True)
+  return announcements
+
+def get_announcements(request):
+  notifications = []
+  if is_trainee(request.user):
+    trainee = trainee_from_user(request.user)
+    notifications += discipline_announcements(trainee) \
+        + server_announcements(trainee) \
```
radicalray: Please line up the indentation to discipline?

ap/announcements/notifications.py
```diff
@@ -0,0 +1,70 @@
...
+        + bible_reading_announcements(trainee) \
+        + request_statuses(trainee)
+  # sort on severity level of message
+  return sorted(notifications, lambda a, b: b[0] - a[0])
+
+def request_statuses(trainee):
+  requests = []
+  requests += IndividualSlip.objects.filter(trainee=trainee, status='F')
```
radicalray: multi-line + here as well.

billyeh:
All right. I have discovered the `chain` function, awesome haha. Apparently more efficient because it's `itertools` too. http://stackoverflow.com/questions/431628/how-to-combine-2-or-more-querysets-in-a-django-view

radicalray:
Super! Could you add this to the wiki? Maybe under django queryset or something to help other Devs in the future. @billyeh

billyeh:
Sure. Done!

PR #426, opened 2016-12-21
fix bugs for trainee scroll and dup message

ksungreborn:
Fix trainee dropdown and duplicate message when internet access approved


billyeh:
Hopefully that's better. The old server also allows, like, GTCA teammates, but do we have an equivalent for that on the new one?

ap/web_access/forms.py
```diff
@@ -35,7 +36,14 @@ class Meta:
 
 class EShepherdingRequest(forms.Form):
     active_trainees = Trainee.objects.select_related().filter(is_active=True)
```
jontey: I don't think you need this line since Trainee Object returns active trainees by default.

jontey:
Yeah.. Gospel trips have not been added yet

ap/web_access/forms.py
billyeh: Yeah, you're right.

billyeh:
I accidentally created the branch for #454 off this branch, so the majority of the changes are already merged into dev. This is basically just a pull request for some one-line code cleanup now.

PR #434, opened 2016-12-23
Ruthnan

ruthnan:
added check for os and selection of the right chromedriver depending. added option to run on Chrome with -d in the absent_trainee_roster test


swenghj:
base should be "automation"

PR #435, opened 2016-12-23
Ruthnan

ruthnan:
merging with automation instead of dev. checking whether the os is mac or linux and appropriately calling the right chromedriver for all test scripts. also, fixed minor bug in absent_trainee_roster.py test script that had browser hardcoded to be firefox.


swenghj:
Great!

PR #436, opened 2016-12-23
Quash startup errors

jontey:
@leonardlan


billyeh:
Hmm this is gonna take some fixing. I did the merge wrong I think.

ap/terms/management/commands/populate_terms.py
```diff
@@ -1,36 +1,35 @@
...
-    def handle(self, *args, **options):
-        self._create_terms()
\ No newline at end of file
+class Command(BaseCommand):
+  # to use: python ap/manage.py populate_terms --settings=ap.settings.dev
+
+  def _create_terms(self):
+    today = date.today()
+    for year in range(2010, date.today().year + 1):
+      fall_start = nth_weekday(2, 0, year, 8)
```
jontey: Isn’t the term calculated from the semi-annual week and counting backwards 20 weeks?

billyeh: That makes sense. I'm not sure whether it was equivalent or not but I changed it to do that because I don't trust dates/times.

ap/apimport/utils.py
```diff
@@ -18,25 +18,35 @@
...
+
 def date_for_day_of_week(date, day):
-    """ returns the date of the specified day in the week identifid by date.
-        For example, if date 4/9/2016 (Saturday) and 0 (Monday) are passed in,
-        4/4/2016 will be returned. """
-    return datetime.combine(date + timedelta(days=day-date.weekday()), time(0,0))
+  """ returns the date of the specified day in the week identified by date.
+      For example, if date 4/9/2016 (Saturday) and 0 (Monday) are passed in,
+      4/4/2016 will be returned. """
+  return date + relativedelta(days=day-date.weekday())
```
jontey: I don't think this gives the desired result even though it is equivalent to the old function

ap/apimport/templates/apimport/term_details.html
```diff
@@ -3,23 +3,20 @@
...
+      {{ term_dates.media }}
+    {% endif %}
     <script>
-        $(document).ready(function() {
-            $('.datepicker').datepicker(); 
-            $(document).on('change','#id_start_date',function(){
-                calculate_dates();
-       });
+      $(document).ready(function() {
+        $(document).on('change','#id_start_date',function(){
```
billyeh: Oops yeah didn't realize someone else might be working on this stuff or I wouldn't have touched it. :(

ap/apimport/utils.py
billyeh: Yeah, gives `date` instead of `datetime` is that what you mean?

jontey: Nvm... Monday is the first day. I used Sunday as my test case and got confused.

requirements/base.txt
```diff
@@ -1,15 +1,15 @@
...
 Django==1.8.1
 psycopg2==2.6
 django-braces==1.8.0
-django-adminactions==0.4
+django-adminactions==1.2
 django-suit==0.2.13
-django-bootstrap3==4.8.2
-django-bootstrap3-datetimepicker==2.2.3
+django-bootstrap3==7.1.0
+django-bootstrap3-datetimepicker-2==2.4.2
```
jontey: are you sure it is datetimepicker-2? I needed to install datetimepicker for my local to run

jontey: interesting... uninstalled and reinstalled the package worked

PR #441, opened 2016-12-25
Ruthnan

ruthnan:
finished writing the HC test case.


swenghj:
could you combine HC role test into absent_trainee_roster.py? I think we can maintain one file including all tests.

Issue #266, opened 2016-06-16
RFID Rolls


rayoowoo:
Spec incoming. If you all think of any edge cases that I haven't covered, or there are any issues, or there are any possible solutions to the issues I listed, please comment!

also, @radicalray I'm not sure where the RFID readers will be - would you happen to know?

rayoowoo:
# Automated Rolls in the Cafeteria

### Introduction

Rolls that take place in the cafeterias will now be taken by swiping badges on an RFID reader, thus replacing handwritten attendance rolls for meal times and study times. 

### Included Rolls

* Breakfast (Tuesday - Saturday) - 7:35am
* Breakfast (Lord's day) - 7:30am
* Lunch (Saturday) - 11:45am
* Dinner (Monday - Thursday, Saturday) - 5:45pm
* Campus & internet gospel team study rolls (Tuesday) - 4:15pm
* Campus & internet gospel team study rolls (Wednesday) - 4:30pm
* Campus & internet gospel team study rolls (Thursday) - 4:45pm
* Afternoon Study Roll (Saturday) - 2:15pm
* Character afternoon study roll (Thursday) - 4:45pm

### Excluded Rolls
Special schedules that have special rolls (e.g. Memorial Day Conference - Monday breakfast rolls) that take place within the cafeteria will be manually set. Rolls that do not occur within the cafeteria are currently not included. These include:

* Evening study rolls (Thursday, Saturday)
* Character study roll on Tuesdays (not regular)
* Children's, YP, and community study rolls (Tuesday - Thursday)
* Friday night study roll
* Class Rolls (includes PSRP)
* Team Rolls
* Service Group Rolls
* House Rolls

### Setup

There is a place on the server where the times for readers are activated for roll-taking. The time of each roll is manually determined:

![automatic meal rolls](https://cloud.githubusercontent.com/assets/17632989/16130980/b17b15e6-33c0-11e6-90ae-6d2cb76bb272.jpg)

For most rolls, tardies will be at the time indicated on the schedule, and absences 15 minutes afterward. For these rolls, the RFID readers will be activated for roll-taking 15 minutes prior to what would be a tardy, and the readers are deactivated at the time of which would be considered an absence. (For example, Tuesday campus study roll is listed as 4:15pm on the schedule. 4:15pm would be considered tardy, 4:30pm considered an absence. The reader would be activated from 4:00pm to 4:30pm.)

Certain rolls do not have a tardy option, only absences. These are:
[...]

PR #442, opened 2016-12-25
Rfid rolls

billyeh:
Not sure how well this follows #266 but added an RFID roll view and some buttons on that page for flipping rolls and removing tardies in the case that we're dismissed too late to swipe within the 15 minute window.


ap/attendance/views.py
```diff
@@ -333,3 +342,58 @@ class AllAttendanceViewSet(BulkModelViewSet):
...
+
+  return HttpResponse('Roll finalized')
+
+@group_required(('attendance_monitors',))
+def rfid_tardy(request, event_id, event_date):
+  event = get_object_or_404(Event, pk=event_id)
+  date = datetime.strptime(event_date, "%Y-%m-%d").date()
+  if not event.monitor == 'RF':
+    return HttpResponse('No event found')
+  rolls = event.roll_set.filter(date=date, status='T')
```
jontey: could you do `event.roll_set.filter(date=date, status='T').delete() ` instead?

jontey: Also need to check if a leaveslip is attached to the roll before deleting.... Perhaps override the delete method in the model?

ap/schedules/models.py
```diff
@@ -123,15 +117,7 @@ def date_for_week(self, week):
 
   # checks for time conflicts between events. Returns True if conflict exists.
   def check_time_conflict(self, event):
-    return (self.end >= event.start) and (event.end >= self.start)
-
-  # gets the week from an absolute date of the current term.
-  def week_from_date(self, date):
```
jontey: Please add back these two functions. Thanks

billyeh: OK. I don't think we need a non-static method, since the static method will work when called with an instance too.

jontey: True.. just keep one of them then

ap/attendance/views.py
billyeh: First comment is correct. I was running into some weird bug before but it seems all right.

I don't think we want to override the delete method that way, especially with what this says http://stackoverflow.com/questions/1534986/how-do-i-override-delete-on-a-model-and-have-it-still-work-with-related-delete. I just added the filters.

PR #443, opened 2016-12-27
Exam grading

ksungreborn:
Fix up bugs related to exam grading, taking exams, and also added functionality to filter by class and term and delete exams.


ap/exams/urls.py
```diff
@@ -1,15 +1,17 @@
...
 from exams import views
-
+from exams.models import Exam
 
 urlpatterns = patterns(
     '',
     url(r'^$', views.ExamTemplateListView.as_view(), name='list'),
     url(r'manage/$', views.ExamTemplateListView.as_view(), {'manage': True}, name='manage'),
     url(r'new/$', views.ExamCreateView.as_view(), name='new'),
+    url(r'manage/delete/(?P<pk>\d+)$', views.ExamDelete.as_view(model=Exam, success_url='/exams/manage/'), name='exam-delete'),
```
jontey: Deleting Exams should be an API call.

ap/exams/templates/exams/exam_graded.html
```diff
@@ -0,0 +1,51 @@
+{% extends "exams/base_exams.html" %}
```
jontey: This page looks similar to ap/exams/templates/exams/exam.html
Could you refactor this part so that we can reduce the amount of duplicate code?

ap/exams/models.py
```diff
@@ -44,6 +46,8 @@ class Exam(models.Model):
     # values assigned for each question.
     total_score = models.DecimalField(max_digits=5, decimal_places=2, default=0.0)
 
+    #passing_percentage = models.IntegerField(validators=[MinValueValidator(0), MaxValueValidator(100)])
```
leonardlan: Removed if not needed.

PR #455, opened 2017-03-18
Frontend reorg

schai6:
Fixed calendar display and submit group slips.


yipingbenjaminliu:
badges has a bug with UI, part of the navbar will be covered
new bitmap image

PR #469, opened 2017-04-02
Fix add partition row

ruthnan:
Jon Tey fixing his own code:
- changed the add partition to make sure it is not empty when it adds a new row
- deleting multiple rows at a time is no longer a problem
- does not add multiple rows at a time


ruthnan:
created a new commit in #494 to not include all of automation code with git push

PR #474, opened 2017-04-11
Add scripts to run automation

swenghj:
* add test setting requirements 
* use the production setting for migrations and running the server 
* all things running after deploy stage in Travis


swenghj:
Setup changed, so another commit will be made

PR #477, opened 2017-04-14
Attendance TA Views

jontey:
Added TA views for leaveslips


ap/aputils/groups.py
```diff
@@ -25,45 +25,55 @@ def app_permissions(group, app_label_list):
...
-    'ypc_monitors',
-    'xb_trainees',
-    'designated_service',
-    'special_projects',
-    'office_support',
-    'badges',
-    'health_office',
-    'kitchen',
+  # Update permissions
+  APPS = [
```
billyeh: It'd be nice to just have this variable in settings/base.py and then we don't have to change it in two places later?

PR #471, opened 2017-04-02
Attendance/Bug fixes

billyeh:
Gonna start a pull request so I can keep track of changes in this branch. 

* fix absent_trainee_roster page so it renders with context processors
* fix static asset links for select2 fields
* add field for attendance monitors to select a trainee to submit rolls for
* working finalize roll button
* address to-dos in #457 as noted there
* add notifications for house requests and rolls

Issue #457, opened 2017-03-18
Attendance Remaining Functionality


czhaocmu:
### 1. Personal Attendance main (Role: Trainee)

**Functionality**

**UI**
- [x] Remove legend button (show the legend by default)
- [ ] Leaveslips under Personal Attendance should have a legend as well, showing what the different states mean
- [x] Remove pixel from between events that show up adjacent on the schedule calendar
- [x] Should the week that the calendar is displaying be dark blue background with blue bar underneath, indicating non-clickable tab? (see attached picture)

![trainee attendance_summary](https://user-images.githubusercontent.com/23424787/29999339-5aa97d0c-8ff9-11e7-8f61-4806ed4a1611.png)

- [x] Add legend to know what the color of events mean
- [x] Add Note: Report information will not be up-to-date until attendance office hours (i.e., when the potential violators list is posted). #471 d610c55a41d71744a8f9872d372511fdfa2c9ec8
- [x] Add title to Attendance page as seen on browser tab (FTTA|Personal Attendance) #471 d610c55a41d71744a8f9872d372511fdfa2c9ec8

Nice to have*
- [ ] Add lines to separate days and times
- [x] Add dropdown to select attendance period 136946f4f83ec43141992edf9c0e618d24575ccc
- [x] Add month of date in calendar header

**Permissions**
- [x] ~~First year trainees~~Non self-attendance trainees should not have roll button #471 48e46ab1ee391fbe2066b52a79872dd64468a0a6
	
### 2. Personal Attendance->Roll (Role: Second Year Trainee)

**Functionality**
- [x] Finalize attendance starting Mon to end of Tues #471 86136070a72eabe66988ac50afc19a76b0206a7f

**UI**
- [x] Add notifications that roll was not finalized. #471 de7f71d937ef6d35fde34343bf9c5a91d9e31dd7

**UX**
- [x] Remove all sessions closes the entire Roll or Leaveslips box?

**Error Handling**
- [x] When no sessions are selected, submit button is not disabled. 
- [x] Should not be able to submit attendance for future and past weeks. #471 969b1d68ec0ae3a69582491f4acc42f9468fd395

**Permissions**
- [x] Add 

### 3. Personal Attendance->Leaveslips (Role: Trainee)

**Functionality**
- [x] Add last slip date for meal out and night out with contact info form. #471 dcba24b81544a1de7f483fb57fee0e0114258766
- [ ] Add submitted late to leaveslips after 48 hours.

**UI**
[...]

PR #490, opened 2017-04-23
Attendance bug fixes and functionality

billyeh:
Comprising #471, attendance-switch-trainee branch, (and #457) and bower-reorg branch. With fixes to announcements, web requests, life studies, attendance, and meal seating


jontey:
When editing a group leaveslip, the events are not selected on the calendar.

ap/attendance/views.py
```diff
@@ -100,6 +100,7 @@ def get_context_data(self, **kwargs):
       event_id = self.request.POST.get('events')
       event = Event.objects.get(id=event_id)
       selected_date = event.date_for_week(int(selected_week))
+      event.date = selected_date
```
jontey: GroupSlips where trainee is not in them but created it is still showing up. 
I used the account that i generated leaveslips for service scheduling and i got a bunch of group leaveslips on my page.

libraries/react/scripts/selectors/selectors.js
```diff
@@ -132,15 +132,10 @@ export const getESRforWeek = createSelector(
...
-      status = categorizeEventStatus(esr.event)
-      if (status.roll || status.slip) {
-        esr.event.status = status
-        evs.push(esr)
-      }
-    });
+    let evs = esrs.map(esr => {
+      esr.event.status = categorizeEventStatus(esr.event)
+      return esr
+    })
```
jontey: What does getGroupSlipsforPeriod do? I dont think the trainee_names is used anywhere

ap/attendance/views.py
```diff
@@ -100,6 +100,7 @@ def get_context_data(self, **kwargs):
       event_id = self.request.POST.get('events')
```
jontey: Lin 59: The trainees selector is generating a lot of db calls. Perhaps consider prefetching groups

ap/attendance/views.py
jontey: Also when editing groupslips, the events are not selected automatically. the user has to select the events again before updating the leaveslip.

ap/schedules/views.py
```diff
@@ -41,67 +40,12 @@ def get_queryset(self):
...
-          schedule.events.add(event)
-      else:
-        schedule = Schedule(trainee=trainee, term=event.term)
-        schedule.save()
-        schedule.events.add(event)
-
-    return super(EventUpdate, self).form_valid(form)
-
-
 class EventDelete(generic.DeleteView):
```
jontey: This view is not used. Could you delete it?

ap/attendance/views.py
jontey: Ok. figured it out. If there is no event on the group slip calendar that falls within the time selected, then no events are selected. It's a small edge case which can be avoided by filling up all the empty slots on the calendar with something

billyeh:
All right. Thanks for reviewing Jon and Leonard!

billyeh:
@jontey Merging this before Saturday this week if there's nothing else to address :)

`multiselect.js` in meal seating static assets should also be moved to npm but https://www.npmjs.com/package/multiselect-two-sides is the package, and it has a very outdated jquery requirement that messes us up. We can address it later by either getting him to upgrade it or doing it ourselves.

PR #494, opened 2017-05-13
Fix partition adding on seating charts

ruthnan:
- changed the add partition to make sure it is not empty when it adds a new row
- deleting multiple rows at a time is no longer a problem
- does not add multiple rows at a time


jontey:
Waiting to merge #490 before merging this branch

billyeh:
Going to close this to keep things on dev clean 👍  the fix partition commits have been cherry-picked to #490. Thanks Ruth and Jon!

Issue #493, opened 2017-05-13
Bug: House Form Requests


swenghj:
## Issues 
- [ ] Fix reports links: they are not shown (maintenance, linens, framing, furniture)
  : these links are for the HCs and designated service trainees 

- [x] Missing fields in submitting requests
   **house number is omitted for implementation purpose, but the house number is not shown when the request is created - fix whether in the form submission or in the form submitted**
  - [x] Maintenance - missing: 1) house number, 2) type, 3) room, 4) priority 
  - [x] Linens - missing: 1) house number, 2) item and reasons should be drop down menu? 3) how about urgent check box?
  - [x] Framing - missing: 1) house number, 2) check box(frame approved by TA brother), 3) change the title of blank, "Frame" to "Desired quote to be in frame: Must be a verse, footnote, ministry quote, or hymn AND must have verbal approval from a TA brother." 4) check box(Is a frame there already?)

billyeh:
Closing because reports are being considered separately

PR #505, opened 2017-06-10
Population/house request fixes

billyeh:
Gonna try to keep the pull requests smaller. Addresses population script issues run into during on-boarding this term and the bugs in #493.


ap/attendance/management/commands/populate_rolls.py
```diff
@@ -13,28 +15,32 @@ def new_roll(trainees=[], am=[]):
...
+                Roll(event=e, trainee=trainee, status=random.choice(['A', trainee_stat]), submitted_by=cur_am, date=Term.get_date(Term.current_term(), week, e.weekday)).save()
 
 class Command(BaseCommand):
     # to use: python ap/manage.py populate_rolls --settings=ap.settings.dev
     def _create_rolls(self):
         # change the array below to get different trainees
         trainees = Trainee.objects.filter(current_term__gt=2)
         am = Trainee.objects.filter(groups__name="attendance_monitors")
+        if not len(am):
+          am_group = Group.objects.get(name='attendance_monitors')
```
jontey: This line returns None on a new database. You need to set some trainees in populate_trainees as attendance monitors

billyeh: That line gets the attendance monitor group, so it works once the groups are populated at least (post-migration script does that). And then the attendance monitors are randomly assigned a few lines down.

PR #508, opened 2017-06-17
Add colors to seating chart seats by term

czhaocmu:
I can change the colors once we finalize them.

### DIFF
--- a/ap/seating/views.py
+++ b/ap/seating/views.py
@@ -6,7 +6,7 @@ from .models import Chart, Seat, Partial
 from terms.models import Term
 from accounts.models import Trainee
 from .serializers import ChartSerializer, SeatSerializer, PartialSerializer, PartialFilter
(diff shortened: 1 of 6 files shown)


ap/seating/views.py
```diff
@@ -6,7 +6,7 @@
 from terms.models import Term
 from accounts.models import Trainee
 from .serializers import ChartSerializer, SeatSerializer, PartialSerializer, PartialFilter
-from accounts.serializers import TraineeSerializer, BasicUserSerializer
+from accounts.serializers import TraineeSerializer, BasicUserSerializer, TraineeRollSerializer
```
jontey: Is BasicUserSerializer still in use? if not we can delete it

czhaocmu: Nope, I removed it.

czhaocmu: termClass css is in seating.css now.

jontey:
Good job @czhaocmu. Merging!

PR #512, opened 2017-06-22
Frontend css

jontey:
CSS color updates
CSS updates for buttons and tables
Deleting unused bootflat files used in their demo site


billyeh:
Gonna merge this one because we can always go back and change CSS.

Issue #478, opened 2017-04-15
Creating Schedules on admin page


allisonfaith:
When creating schedules on the admin page, there is no option to select first year or second year trainees. And no way to sort the list of trainees by category, only by typing a search. 
This will be very difficult when making more schedules in the future.

PR #513, opened 2017-06-22
Upgrade select2

jontey:
Upgrading django-select2
Working on #478


billyeh:
@schai6 You ran into the new version of django_selecte2 requesting select2 javascript from a CDN, right? 

That's what [the repo](https://github.com/applegrew/django-select2) says it does now. We'll have to see if we can change it to use some local assets (probably using the import from bundles/index.js since we already use it for the CSS), and tell it not to look for the assets online http://django-select2.readthedocs.io/en/latest/_modules/django_select2/conf.html?highlight=cdn

jontey:
I think the problem with extending the select2 on jQuery is because django admin already loads a different version of jQuery. what we are extending is not $ but jQuery... i think the only way to deal with that would be to prevent django admin from loading their own jQuery and loading our own bundle

billyeh:
You're right Jon. Works in 10e7448 in #521

billyeh:
This is all mixed up with #521 now so gonna close.

Issue #476, opened 2017-04-14
Leaveslips TA Views - TODO


jontey:
Some thoughts to improve leaveslips display for TAs

- [x] Change color for calendar display to be more obvious which events are selected #518
- [x] Old Server shows trainee's house and team
- [x] If type is house night out or meal out, show last approved night out for trainee #518
- [x] Add TA private remarks.. (for between TAs) #518
- [x] Individual Leaveslips should show trainee's attendance report #518

PR #518, opened 2017-06-28
476/ta leaveslip views

iamsitting:
#476


ap/leaveslips/templates/leaveslips/leaveslip_update.html
```diff
@@ -27,20 +27,55 @@
...
 
+    for (var i=0; i < Events.length; i++){
+      for (var j=0; j< SelectedEvents.length; j++){
+        if(!Events[i].start.localeCompare(SelectedEvents[j].start)){
+          Events[i].color = LeaveslipColor;
+	}
+      }
+      for (var j=0; j< AttendanceRecord.length; j++){
+	if(!Events[i].start.localeCompare(AttendanceRecord[j].start)){
+	  if(AttendanceRecord[j].att == 'T'){
```
jontey: Could you simplify the logic here? 
Eg.
```
var calendar_color = {
  'T': '#F69D74',
  ...
}
```

then over here you can just do 
`Events[i].color = calendar_color[AttendanceRecord[j].att];`

ap/leaveslips/views.py
```diff
@@ -36,9 +38,36 @@ def get_context_data(self, **kwargs):
...
+      # end getting attendance record
+      
+      ctx['attendance_record'] = json.dumps(attendance_record)
       ctx['events'] = leaveslip.trainee.events_in_date_range(start_date, end_date)
       ctx['start_date'] = start_date
       ctx['end_date'] = end_date + timedelta(1)
+      ctx['selected'] = leaveslip.events
+      if (leaveslip.type == 'MEAL' or leaveslip.type == 'NIGHT'):
+        last_leaveslips = IndividualSlip.objects.exclude(id=leaveslip.id).order_by('-rolls__date').filter(trainee=leaveslip.trainee, type=leaveslip.type, status='A')
+        if len(last_leaveslips) > 0:
```
jontey: i think it is more efficient to use last_leaveslips.exists()
https://docs.djangoproject.com/en/1.11/ref/models/querysets/#django.db.models.query.QuerySet.exists

ap/leaveslips/views.py
```diff
@@ -36,9 +38,36 @@ def get_context_data(self, **kwargs):
...
+          last_leaveslip_date = last_leaveslips[0].events[0].date
```
jontey: actually since you only want the first leaveslip that fits the parameters you should use `last_leaveslips.first()`

ap/leaveslips/views.py
jontey: Disregard this comment

ap/leaveslips/views.py
```diff
@@ -36,9 +38,35 @@ def get_context_data(self, **kwargs):
...
       end_date = Term.current_term().enddate_of_period(periods[-1])
+      
+      # getting attendance record
+      rolls = leaveslip.trainee.rolls.exclude(status='P')
+      slips = IndividualSlip.objects.exclude(id=leaveslip.id).filter(trainee=leaveslip.trainee, status='A')
+      slips = list(chain(slips, GroupSlip.objects.filter(trainee=leaveslip.trainee, status='A')))
+      attendance_record = []
+      ev_check = []
+      for slip in slips:
+          for ev in slip.events:
```
jontey: This line errors out if trainee has a groupslip that is approved

ap/leaveslips/views.py
```diff
@@ -36,9 +38,35 @@ def get_context_data(self, **kwargs):
     if len(periods) > 0:
       start_date = Term.current_term().startdate_of_period(periods[0])
       end_date = Term.current_term().enddate_of_period(periods[-1])
+      
+      # getting attendance record
+      rolls = leaveslip.trainee.rolls.exclude(status='P')
+      slips = IndividualSlip.objects.exclude(id=leaveslip.id).filter(trainee=leaveslip.trainee, status='A')
+      slips = list(chain(slips, GroupSlip.objects.filter(trainee=leaveslip.trainee, status='A')))
```
jontey: Also this filter is wrong. It should be `trianees_in=[leaveslip.trainee]` not `trainee=leaveslip.trainee`
the trainee field in groupslips is based on the person who created the leaveslip not the person for whom the leaveslip is for.

ap/leaveslips/views.py
```diff
@@ -21,24 +23,81 @@
...
 
 class IndividualSlipUpdate(GroupRequiredMixin, generic.UpdateView):
   model = IndividualSlip
   group_required = ['administration']
   template_name = 'leaveslips/individual_update.html'
   form_class = IndividualSlipForm
   context_object_name = 'leaveslip'
+  
+  
+  def get_attendance_record(self, leaveslip):
```
billyeh: This is a super useful function. We should move it out somewhere, like the Trainee model, so that other modules can use it.

billyeh: Moved it into `accounts/models.py` so we can make sure that one works and get rid of this one. 

Also, I don't quite see why we need to exclude the current leaveslip so I removed that. Otherwise, should be exactly the same.

billyeh:
Looks good. Merging this in!

PR #523, opened 2017-06-30
Fix event time

ruthnan:
event date small issue


billyeh:
Relevant changes in #524

PR #525, opened 2017-06-30
Class table roll

ruthnan:
ordered by date time vs priority only


billyeh:
Commit merged into #524 by Benji

PR #530, opened 2017-07-26
Room Reservations

iamsitting:
Room Reservations app:
Trainees can requests rooms and edit pending requests
TAs can approve/deny requests and view reservation schedule


billyeh:
Good job, thanks for putting in interim work, Carlos!! 😁
We are now like, 👌 this close to having the "required" modules done. This term's gonna be exciting.

iamsitting:
Just realized I forgot to handle TAs making room reservations. Current model only handles trainees making requests.

billyeh:
Yeah, it could be as simple as linking the trainee request page and making it auto-approved and removing making that trainee field something else.

iamsitting:
One more problem: Need to filter requests by trainee for "submitted requests" table.

billyeh:
Wait, for the requester can we just use a user field? User should be able to be either a trainee or TA. Then automatically fill it in before. Haha because what if people don't put their full names the same as it is in the server?

iamsitting:
Oh yeah, forgot about User class...

billyeh:
What do you mean by filtering requests by trainee?

iamsitting:
"Term" reservations are only reflected in the UI, not in the DB. Not sure if I need to auto-generate individual reservations for the TV UI.
I.e. "Term" reservation made on week 10, would generate 10 reservations.

billyeh:
Yeah, I see how you're doing it dynamically. That's great. We can worry about the UI later I think.

billyeh:
Haha. Now I see the little disagreement we had in efec2a8b541817c0158e1b5749f5499a27b26cce. Do we just read the schedule table differently? 😆

iamsitting:
Ahh I guess so. I think the current UI uses 2 blocks to indicate a half hour reservation and 3 for 1 hour. If we do this way, we'd have to force a 2 block min selection.

ap/room_reservations/templates/room_reservations/request_list.html
```diff
@@ -0,0 +1,82 @@
...
+              </a>
+            {% elif reservation.status != 'A' %}
+              <a href="{{ reservation.get_update_url }}">
+                <button type="button" title="Approve" class="btn btn-primary">
+                  <span class="glyphicon glyphicon-edit" aria-hidden="true"></span>
+                </button>
+              </a>
+              <a href="{{ reservation.get_delete_url }}">
+                <button class="btn btn-danger delete_request_button" type="submit" title="Delete"/>
+                <span class="glyphicon glyphicon-trash"></span>
```
billyeh: spacing

PR #542, opened 2017-08-28
Population script fixes

iamsitting:
Fixed terms bug and population order

### DIFF
--- a/ap/accounts/management/commands/populate.py
+++ b/ap/accounts/management/commands/populate.py
@@ -4,10 +4,10 @@ from django.core.management import call_command
 class Command(BaseCommand):
   def handle(self, *args, **options):
     call_command('populate_terms')
+    call_command('populate_schedules')
     call_command('populate_events')
     call_command('populate_trainees')
     call_command('populate_tas')
     call_command('populate_rolls')
     call_command('populate_services')
-    call_command('populate_schedules')
     print('You may now want to manage permissions/groups or create a superuser: ./manage.py createsuperuser')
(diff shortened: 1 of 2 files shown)


ap/apimport/utils.py
```diff
@@ -55,6 +55,13 @@ def term_start_date_from_semiannual(season, year):
   # return date of 19 weeks previous-- one week for semi-annual
   return datetime.combine(seed_date + timedelta(weeks=-19, days=0), time(0, 0))
 
+def term_end_date_from_semiannual(season, year):
```
billyeh: Looks like this disappeared in 50932eae7ec14e953381497a99cbd8df95c9e175. What a monster merge.

ap/accounts/management/commands/populate.py
```diff
@@ -4,10 +4,10 @@
 class Command(BaseCommand):
   def handle(self, *args, **options):
     call_command('populate_terms')
+    call_command('populate_schedules')
```
billyeh: Sorry, why do we want to move schedules up? I remember it didn't work if it was above trainees and events since we add those to the created schedules.

ap/apimport/utils.py
jontey: I might have deleted it because it wasn't used?

ap/accounts/management/commands/populate.py
iamsitting: I was getting some NoneType object errors running the populate script and moving schedules fixed it.

ap/apimport/utils.py
iamsitting: It was being used by populate_terms

PR #555, opened 2017-09-16
Class rolls table

billyeh:


billyeh:
Haha I don't know if we need a review for this. I just don't like pressing the big red override button 🙁

Issue #559, opened 2017-09-21
Class attendance rolls to seating chart not correlating


ksungreborn:
The class attendance rolls to update trainees' attendance does not correlate with seating chart (some people are missing from the chart in class attendance roll view)

ruthnan:
I think this is because the schedule that the events are assigned to don't have a term associated with it/we can't see anything on the schedule-side on the admin page. related to issue #562

jontey:
Actually the issue is that the event has last term's seating chart associated with it.
![screenshot-2017-9-23 change event ftta admin](https://user-images.githubusercontent.com/1665131/30779383-a40c7c0a-a0a3-11e7-90ac-6dcee54f5821.png)

Seating chart should be Fall 2017, not Spring 2017

jontey:
nvm.. that didn't change anything

ruthnan:
Yes, I tried with various classes already...

jontey:
Fixed in #579

PR #560, opened 2017-09-21
switch tardy and uniform tardy order

ksungreborn:


ksungreborn:
Actually, not working as required yet. Let me keep working on it

billyeh:
Wonder if we can fix #559 in here too?

Issue #558, opened 2017-09-20
TA Leaveslip views Bug


jontey:
TA Leaveslip shows leave slips from past terms

ruthnan:
following @finben's style on tracking bugs in Projects>>Testing

Issue #565, opened 2017-09-24
Unicode Issue for Cities


jontey:
Can't figure out which widget is giving the issue when loading cities with unicode characters in Trainee view

jontey:
I believe that the issue is in ModelSelect2MultipleWidget

billyeh:
Cannot reproduce on live server.

jontey:
Found the issue.. It is in the unicode function for Address in aputils/models

PR #567, opened 2017-09-26
Bug 558/ta leaveslip bug

jontey:
Update object manager for leaveslips to only show current term leaveslips.


jontey:
Oops. not ready yet. Still need to work on #565

ap/leaveslips/models.py
```diff
@@ -160,18 +175,34 @@ def get_ta_update_url(self):
   def get_update_url(self):
     return reverse('leaveslips:individual-update', kwargs={'pk': self.id})
 
+
+class GroupSlipManager(models.Manager):
```
billyeh: What happens if we just use the same manager for both classes? It's the same code.

jontey: Group Slips have start and end times, for IndividualSlips we use rolls

billyeh: Oh yeah.

jontey:
I think it's good for review
Fixes #558

billyeh:
Yeah, looks good.

PR #570, opened 2017-09-28
Bug 568/announcements not able to be created

billyeh:
Need to use `ModelMultipleChoiceField` instead of `ModelChoiceField` and add `save_m2m` for the created object.

### DIFF
--- a/ap/announcements/views.py
+++ b/ap/announcements/views.py
@@ -34,6 +34,7 @@ class AnnouncementRequest(generic.edit.CreateView):
     req = form.save(commit=False)
     req.trainee_author = trainee_from_user(self.request.user)
     req.save()
+    form.save_m2m()
     return super(AnnouncementRequest, self).form_valid(form)
 
 class AnnouncementRequestList(generic.ListView):
--- a/ap/announcements/views.py
+++ b/ap/announcements/views.py
@@ -32,7 +32,7 @@ class AnnouncementRequest(generic.edit.CreateView):
 
   def form_valid(self, form):
     req = form.save(commit=False)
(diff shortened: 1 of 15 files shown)


ap/announcements/views.py
```diff
@@ -34,6 +34,7 @@ def form_valid(self, form):
     req = form.save(commit=False)
     req.trainee_author = trainee_from_user(self.request.user)
     req.save()
+    form.save_m2m()
```
jontey: +1

ksungreborn:
I notice announcements created by trainees have no way to view it's status and no way of being approved as of now.

jontey:
Is view announcements only for TAs?

billyeh:
There's the view announcements page which is supposed to show which announcements they're supposed to announce in class that day. It should be a printable report, I think, that only TAs can view?

jontey:
Ok. Then the link on the smart menu should be changed to TA only

billyeh:
It is. https://github.com/attendanceproject/djattendance/blob/40cd49e2febf36ba90a0e726ed90f29c093c25d6/ap/aputils/templatetags/smart_menu.py#L100 

But one thing we can do is move the announcement-request-list link to common, rather than being in both trainee and TA.

jontey:
Didn't do a pull before I started reviewing :S

ksungreborn:
@billyeh I wanted to put announcement-request-list in common, but I felt the name for the tab for TA should be 'Create/Approve Announcements' and not just 'Create Announcements', because TA's will be approving not just creating.

billyeh:
Good point!

PR #595, opened 2017-10-20
Upgrade ortools

billyeh:
Ortools has been published to pypi! The one function we use (pywrapgraph) works still. We can finally get rid of this dependencies folder :)


billyeh:
@jontey what do you think? We can get rid of this step of install for the onboarding next week.

jontey:
I think we are using a specific version of ortools v5.1.4045 that is not found in the pypi repo

billyeh:
Oh yeah, I mean, that used to be the most up-to-date version. They only recently got it up on PyPi https://github.com/google/or-tools/releases/tag/v6.4 

It's actually been installable from pip from v5.1.4045 on Mac (just have to upgrade pip sometimes if it's out of date). But yeah I thought the reason wasn't so much about the specific version as it was that they just didn't have linux distros up on pip?

PR #597, opened 2017-10-20
Services Worker Import

jontey:
Fix import/update worker list for services


ap/services/admin.py
```diff
@@ -87,8 +87,10 @@ def get_urls(self):
 
   def import_trainees(self, request):
     from django.http import HttpResponseRedirect
+    # Deletes all worker instances of inactive trainees
+    Worker.objects.filter(trainee__is_active=False).delete()
     # Gets all active trainees and check if any do not have worker affilliation, if not, create one
```
ruthnan: sp. affiliation not affiliation

PR #598, opened 2017-10-20
fix bug with repeated bottom border

jontey:
@billyeh  Do you still want to merge this?


billyeh:
Actually I found a better way to do it and it's on dev already, I think I used another branch to do it. Sorry.

PR #599, opened 2017-10-20
HC Permissions

ruthnan:
added permissions for only HCs to view house rolls and be able to report trainees in the absent trainee roster


iamsitting:
@ruthnan could you merge this with #600 and add permissions to the hc app?

ruthnan:
Done. They're now in #600, and I added permissions for your app, too. There was this line,
>  #if specific_perm_item.permission in context['perms']:
which I think doesn't work for special cases, because the permission is usually something like 'app.random_name' and the context doesn't have that line in there... except for some reason it worked for Absent Trainee Roster...

Issue #601, opened 2017-10-28
Graduation Module


yipingbenjaminliu:
Replicate the old attendance server.

yipingbenjaminliu:
Set Model:
Class Testimony
field, top_experience, textfield
field, encouragement, textfield
field, overarching_burden, textfield
field, highlights, textfield

Class Admin
field, remembrance_text, textfield,
field, remembrance_reference, textfield
field, grad_invitations, smallintergerfield
field, grad_dvd, smallintergerfield

Class Consideration
field, attend_XB, choicefield:
 Yes, I would like to go to FTTA-XB
 Open and considering going to FTTA-XB
 No, it's not likely I will go to FTTA-XB
 other
field, fellowshipped, choicefield:
 Yes
 No
 other
field, financial, choicefield:
 All finances have been fellowshipped and are taken care of
 I may need to fellowship for help with part of the finances
 I may need to fellowship for help with all of the finances
 other
field, consideration_plan, textfield
field, comments, textfield

yipingbenjaminliu:
#602 clear

PR #602, opened 2017-11-02
601/graduation

iamsitting:
Graduation app:
Allows trainees to take Grad surveys

TODO:
~Stat tracker (# of responses, etc.)
Speaking trainees selectform~
Add permissions


iamsitting:
@yipingbenjaminliu you can play around with this app too... Before I keep going. 😃

yipingbenjaminliu:
I updated #601 with the models for graduation. Please adjust accordingly and set it as a main navbar beside Attendance-Discipline-Request-Misc(-Grad). Set the current survey work aside for now, it may come in real handy in the future. It may be good to just comment off the code. So testimony and consideration would both be a form on it's own. For remembrance, please make the two remembrance field a form, and the other two grad field in two separate forms.

yipingbenjaminliu:
fixes for grad admin:
- only if choice is yes for show status, display due date // if choice for show status is no or show only, remove due date display

fixes for all other pages:
- only allow user to edit the field when choice for shows status is yes and before the due date
- if past the due date, make all fields read only // if show only, make all field read only

yipingbenjaminliu:
@iamsitting , fixes to model
Survey:
- remove show_status, due_date
- add a foreign key relation to Grad_admin
- relationship: Survey to Grad_admin = many to one

Grad_admin:
- add a one to one relation to term
- set that default to current term upon creation of grad_admin

am I making sense?

ap/graduation/models.py
```diff
@@ -0,0 +1,232 @@
...
+class Survey(models.Model):
+
+  SHOW_CHOICES = (
+    ('NO', 'no'),
+    ('YES', 'yes'),
+    ('SHOW', 'show only')
+  )
+
+  # grad admin controls the survey
+  grad_admin = models.ForeignKey(GradAdmin)
```
yipingbenjaminliu: makeallmigrations.py throws this response
You are trying to add a non-nullable field 'grad_admin' to website without a default; we can't do that (the database needs something to populate existing rows).
is there someway to set the model so the default grad_admin is the one for the current term

ap/graduation/models.py
```diff
@@ -0,0 +1,232 @@
...
+      'Outline': self.outline_show_status,
+      'Misc': self.misc_show_status,
+    }
+    return SHOW_STATUS_OF[survey_name]
+
+  def get_absolute_url(self):
+    return reverse('graduation:grad-admin')
+
+
+class Survey(models.Model):
```
yipingbenjaminliu: actually, is it necessary to split between the GradAdmin object and the Survey object? Or could we just combine the two?

yipingbenjaminliu:
also, for the current model design, when I run makeallmigrations, it asks for default values for each component of the survey

ap/graduation/models.py
iamsitting: Necessary... it depends. I separated them to simplify the front-end for the grad admin page (taking advantage of modelForm features). And instead of controlling x amount of due dates/show statuses, you just control one (due dates and show status should be the same for all trainees).

iamsitting:
Nuke DB, makeallmigrations, migrate worked for me with latest commit 😆

ruthnan:
@iamsitting sorry, meant to comment on this one... the permissions aren't working here because the fix is in hc-forms, I think...

PR #607, opened 2017-11-14
fix graded view to see scores and fill in the blank bug

ksungreborn:
Correct score for each section can be seen in graded view.
Fill-in-the-blank bug with more than 2 answers has been fixed

### DIFF
--- a/ap/exams/utils.py
+++ b/ap/exams/utils.py
@@ -207,7 +207,6 @@ def save_exam_creation(request, pk):
       total_score += question_point
       options = ""
       answer = ""
-
       if section_type == "MC":
         for k, v in question.items():
           if 'question-option-' in k:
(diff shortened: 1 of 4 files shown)


ap/exams/utils.py
```diff
@@ -224,7 +223,7 @@ def save_exam_creation(request, pk):
       elif section_type == "TF":
         answer = question["answer"]
       elif section_type == "FB":
-        for k, v in question.items():
+        for k, v in sorted(question.iteritems()):
```
jontey: Any reason you are using iteritems here? 

It seems that the function has been removed in Python 3. Maybe revert it back to items() for future compatibility? 
https://stackoverflow.com/questions/13998492/iteritems-in-python

billyeh: 👍 good idea

ksungreborn:
Looks good to me, brothers! I don't have the ability to merge, can one of you?

Issue #605, opened 2017-11-09
Exam auto-grading


ksungreborn:
- [x] Not sure what changes were made to the exam auto-grading part, but after making a test with one of each type of question and taking the test, the exam score is incorrect for what seems to be more than one category of question.

- [x] Also I had previously set the default to: after a trainee finalizes an exam, on the TA view, they must finalize the exam score on their side because of any essay portions that need to be read and graded (for essay questions the score would be 0 until a TA grades it). Now it appears the exam score is already finalized on the TA view with the exam score being '0', which should not be right? Unless we want the trainee to see their score without the essay portion being graded right after they take it? In that case the TA needs to ensure they unfinalize every exam and score the essay portion?

billyeh:
Related: I think for exams we should have a way for trainees to peer-grade essays because that's what we usually do in class.

ksungreborn:
Good idea. I can work on that.

ksungreborn:
Also just noticed a few more bugs:
- looks like 'Add Question' button only works for the first question, after that no more questions are generated for a section.
- total exam score for a test still looks off when I do "VIEW GRADED RESPONSES"

billyeh:
#608 addressed remaining issues

PR #608, opened 2017-11-16
Don't mark auto-graded exam as graded if it has essay questions

billyeh:
And still allow trainee to see immediate feedback for auto-grade. 

Addresses second point in #605

### DIFF
--- a/ap/exams/templates/exams/exam_form.html
+++ b/ap/exams/templates/exams/exam_form.html
@@ -21,20 +21,13 @@
   }
 
   function GetDynamicTextBox(value) {
-    switch (value) {
-      {% for question_type in section_templates %}
-      case "{{question_type}}":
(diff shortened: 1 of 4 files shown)


ksungreborn:
We probably don't want trainees to be able to "VIEW GRADED RESPONSES" before they finish taking the exam right? The answers are on that view =)

billyeh:
Yeah haha good point. I misread the if statements, just need to make sure they've finalized it too.

ksungreborn:
Sorry, just noticed on my branch 'exams-test' the score for 'VIEW GRADED EXAM' looks correct, so I believe only on this branch it shows as '0' regardless of what score the trainee should've gotten.

billyeh:
Works on my branch. We can see if it's reproducable later.

PR #638, opened 2017-12-05
population scripts adjustments for demonstration purposes

yipingbenjaminliu:


ap/accounts/management/commands/populate.py
```diff
@@ -3,15 +3,25 @@
...
     call_command('populate_rolls')
     call_command('populate_services')
-    call_command('populate_biblereading')
-    call_command('populate_books')
-    call_command('populate_houses')
+    call_command('populate_biblereading')    
+    
+
+    # not sure what this model is currently being used for...
+    # call_command('populate_books')
```
billyeh: This is for life study submission. You need a book and chapter for it. E.g. LS Genesis message 10.

Also for syllabus and library would be future applications.

Issue #238, opened 2016-06-02
Import remaining work


haileyl:
- Major logging work
- Permissions
- UNIT TESTS GALORE 
- look into gisgraphy

ricefield:
+1 for unit tests!

jontey:
@haileyl Do you have a progress report for this module?

jontey:
Left TODO for P1. 

- Permissions setting (now only for HCs)
- Schedules import

iamsitting:
Let's leave this open. This one is kind of important. Also related to #722

PR #653, opened 2017-12-14
Schedules import

iamsitting:
Schedules import (#238) is done with schedules.utils.split_schedule.

It seems like it does what we want it to do:

- Mid Term
  - Splits schedule at current week
  - Assigns trainees to new schedule
  - Clears trainees from old schedule
  - Split schedule won't be imported to next term

- New Term
  - New term is created
  - Schedules are imported to next term
  - Clears trainees from imported schedule

Not sure what else to test for?


ap/schedules/models.py
```diff
@@ -380,10 +377,9 @@ def assign_trainees_to_schedule(self):
     else:
       week = term.term_week_of_date(datetime.today().date())
 
-    weeks = eval(self.weeks)
+    weeks_set = Set(eval(self.weeks))
```
billyeh: This is now a built-in data type: just use `set(eval(self.weeks))`. No need for `from sets import Set` either.

ap/apimport/utils.py
```diff
@@ -690,7 +687,7 @@ def migrate_schedules():
   schedule_set.extend(schedules)
 
   for schedule in schedule_set:
-    s_new = migrate_schedule(schedule)
+    migrate_schedule(schedule)
```
jontey: migrate_schedules function should be smarter to determine if import is run during midterm or not. 

if it is not midterm, then it should copy all schedules from previous terms. Also it has to check if that schedule has already been imported. (Sometimes we run import function multiple times during the first week).

if it is midterm, (this means that we are running import after teams have been finalized, aka after draft night). then we need to run split_schedule ONLY if the trainee set gets changed, meaning new trainees have been added/removed from the team. 

i believe that this is what's not implemented yet.

ruthnan:
It works if we merge hc-forms, because I have a fix in there for the permissions....

iamsitting:
@yipingbenjaminliu If you want to look at this it would be good. There are a few items that need to be addressed or tested:
- [ ] MapZen to Google Api #722 
- [ ] Schedules import

ap/schedules/models.py
```diff
@@ -359,6 +359,41 @@ def get_roll_table_by_type_in_weeks(trainees, type, weeks, team=None):
     w_tb = EventUtils.collapse_priority_event_trainee_table(weeks, schedules, t_set)
     return EventUtils.flip_roll_list(EventUtils.export_typed_ordered_roll_list(w_tb, type))
 
+  def save(self, *args, **kwargs):
+    self.save_query_filter()
+    super(Schedule, self).save(*args, **kwargs)
+
+  def save_query_filter(self):
```
iamsitting: I'm not sure if we still need this...

PR #693, opened 2017-12-24
add trainig assistant group just for tas

yipingbenjaminliu:


jontey:
What was the reason for this change?

yipingbenjaminliu:
In consideration for the ta leaveslip, along for some future apps, it may be good to have a ta group for permissions rather than trying to set the permission by checking the user status.

PR #790, opened 2018-03-01
logic check for first year attendance finalization

yipingbenjaminliu:


ap/announcements/notifications.py
```diff
@@ -92,6 +92,9 @@ def attendance_announcements(trainee):
   today = datetime.date.today()
   term = Term.current_term()
   week = term.term_week_of_date(today)
-  weeks = map(str, filter(lambda w: not term.is_attendance_finalized(w, trainee), range(week)))
+  if trainee.current_term > 2:
```
jontey: would it be clearer to use trainee.self_attendance instead?

ap/terms/models.py
```diff
@@ -204,9 +204,10 @@ def is_attendance_finalized(self, week, trainee):
     term = self.current_term()
     week_start = term.startdate_of_week(week)
     week_end = term.enddate_of_week(week)
-    if not trainee.rolls.filter(date__lt=week_end, date__gt=week_start, finalized=True).exists():
+    if trainee.rolls.filter(date__lte=week_end, date__gte=week_start, finalized=True).count() > 0:
```
billyeh: I'm not sure this is the change we wanted?

yipingbenjaminliu: I forgot what I was trying to do here...the logic was funny. Feel free to fix it up

Issue #801, opened 2018-03-03
Class Rolls Needs Study Rolls


iamsitting:
- [x] Add Thursday and Saturday Study Rolls 1b5f772d8839693d4ff561a52d2a3fd57491d518
- [x] Current roll shows only Greek II (needs to show German, Greek I, and Character, etc.)
- [x] Freeze the top few rows for convenient roll input.
- [x] Option to order by last name
- [x] When hovering over row, the name is opaquely highlighted (can't see the name)

![screen shot 2018-03-03 at 1 10 29 pm](https://user-images.githubusercontent.com/8170930/36939364-74e53a7e-1ee4-11e8-9773-8864907592df.png)

ruthnan:
Greek I, German, and Character just need people in the class. So AMs need to assign the students. When I put a trainee in Greek I, the class shows up in the Class Table.

yipingbenjaminliu:
who's working on this?

iamsitting:
It's up for grabs.

yipingbenjaminliu:
I'll take it

yipingbenjaminliu:
@iamsitting , anything else pressing over there?

iamsitting:
@yipingbenjaminliu not really, we have jontey working on service scheduling.

yipingbenjaminliu:
Hallelujah Jesus

yipingbenjaminliu:
#831 should solve this.

PR #831, opened 2018-03-07
801/rolls view

yipingbenjaminliu:
to address all the issues in #801


yipingbenjaminliu:
@ruthnan pull this and see if it matches all the criteria to fix #801

billyeh:
- They probably want it to re-sort when clicking the last/first name button
- The header part with the events/times/rolls doesn't scroll to the right even when the inputs below do

billyeh:
Looks good!

Issue #783, opened 2018-02-28
When an event is changed leave slips corresponding to it no longer apply


billyeh:
Example: this pre-training GOW and Apostles' Teaching were switched.

https://github.com/attendanceproject/djattendance/blob/89e41094b4340b7bfd49bfb7d3449c8daf9b16de/libraries/react/scripts/selectors/selectors.js#L98-L106

Whether the leave slip applies depends on the date of the event.

davidhtien:
Is this still an issue?

yipingbenjaminliu:
yes. It will be address when the front end for schedule changes are built

yipingbenjaminliu:
#1306

PR #842, opened 2018-03-10
Leaveslip PR for issue 828

iamsitting:


yipingbenjaminliu:
94b37d7
screenshot from 2018-03-14 19-57-05
@caricchow @xuefeizheng @isaactsou does this make the selected event more obvious? as least temporarily before we come down with a fix...

billyeh:
Lol it's too ugly, we don't need a temporary fix because it's not that critical.

I realized there was a bug with the event not even being less opaque because of #783 which is probably really what the TAs were seeing, I think the opacity thing is actually pretty visible to them.

billyeh:
This guy is ready to merge/for review.

PR #847, opened 2018-03-11
Leaveslip bootstrap alerts

davidhtien:


yipingbenjaminliu:
https://github.com/attendanceproject/djattendance/blob/2c7f90e1dc4584b075889831ae9e7092c31910ba/ap/templates/base.html#L49

yipingbenjaminliu:
they currently exist in the base template

https://github.com/attendanceproject/djattendance/blob/2c7f90e1dc4584b075889831ae9e7092c31910ba/ap/templates/base.html#L23-L49

https://github.com/attendanceproject/djattendance/blob/2c7f90e1dc4584b075889831ae9e7092c31910ba/ap/templates/base.html#L118-L124

yipingbenjaminliu:
@billyeh , what's the best way to go forward with this?

billyeh:
Yeah, surprisingly (or not) the global stuff works in React. 🤔  

Still think we need to refactor that stuff though to use probably Javascript classes and be encapsulated. The saved notification isn't that noticeable, we have announcements notifications already, we can use bootstrap alerts, and we need different options for text. "Saved" doesn't make as much sense to me as "submitted" for leave slips and rolls, for example.

PR #867, opened 2018-03-17
made buttons bigger

samvictor232:
approve/fellowship/deny/save buttons are bigger


billyeh:
Maybe we can try just using the `btn-large` class? https://getbootstrap.com/docs/3.3/css/#buttons-sizes

billyeh:
This ought to fix the bug on Safari/iPad with people's event times on the React calendar being 7 hours early.

PR #871, opened 2018-03-17
Leave slip ta comments

joeyu23:
@iamsitting


joeyu23:
@billyeh Nice changes. Didn't consider the group slips, so good call on that.

PR #877, opened 2018-03-18
hashing selected schedule events

bdou:
also homepage cards have colored numbers and tardies are now $tan.


yipingbenjaminliu:
good work!

PR #884, opened 2018-03-22
text robustness

yipingbenjaminliu:


billyeh:
Lol I'll merge 5193863e9b380d8bdb1a34c0d806a3f6d0002579 from #878 should fix this once and for all.

The idea is we have a prop `requester_name` to always use if we want a string, no errors.

PR #930, opened 2018-03-27
This is to increase the duration of the flash acknowledgements

finben:
As requested by TA's. EX when a trainee submits a leave slip


billyeh:
Nice 👍

billyeh:
Ideally I think we reuse the current notification bell thing or put some permanent alert at the top such as https://getbootstrap.com/docs/3.3/components/#alerts-dismissible but probably lower priority/pending UX review.

PR #940, opened 2018-03-29
This is to change default value of TA to null

finben:
Currently the TA is linked to the user by default. However, this is resulting in inaccurate leaveslips when trainees rush through. So the requested change is to make this field default to null


finben:
Doesn't seem to to work on my local for group leave slip as these are still defaulting to the submitters TA. I am guessing because there is a state change from initial... Also, if you go from group back to personal it also will go back to defaulting to the correct TA

billyeh:
@finben works now if you wanna check it and we can merge.

billyeh:
Whoops I committed to dev 😢 sorry

Issue #863, opened 2018-03-14
When service schedulers update designated services it should update groups


billyeh:
for permissions, to see links and stuff.

ruthnan:
i'll do it

JoannaRumbley:
Also, Designated services are not listing everyone (ex: only 2 brothers – Greg Lee and Ryan Holt – are listed for transportation) http://ap.ftta.lan/services/designated_services_viewer

ruthnan:
closed in #958

PR #958, opened 2018-03-31
Designated service groups

ruthnan:
Addressing #863 #960


ap/services/admin.py
```diff
@@ -296,6 +296,8 @@ def save_related(self, request, form, formsets, change):
     workers = instance.workers.all()
     trainees = Trainee.objects.filter(worker__in=workers)
     for g in permission_groups:
+      if g.name not in ['training_assistant', 'saturday_training_assistant', 'grad_committee', 'HC', 'team_monitors', 'ypc_monitors', 'designated_service', 'PSRP_facilitator']:
```
billyeh: @ruthnan actually this should be fine without the `if` statement, right? It only empties and re-adds `instance.permission_groups` so updating this for like Chair Stars won't erase the TAs, right?

PR #968, opened 2018-04-02
Small changes to solve several requests

finben:


billyeh:
that's a nice yellow.

PR #971, opened 2018-04-03
Attendance report

ksungreborn:
Attendance report generation to merge


ap/reports/views.py
```diff
@@ -0,0 +1,198 @@
...
+    filtered_rolls = Roll.objects.filter(trainee__in=filtered_trainees, date__in=date_list).exclude(status='P')
+    
+    items_for_query = data['general_report']
+
+    for trainee in filtered_trainees:
+      rtn_data[trainee.full_name] = {}
+      for item in items_for_query:
+        rtn_data[trainee.full_name][item] = 0
+    
+    for roll in filtered_rolls:
```
yipingbenjaminliu: maybe instead of running through all the if conditions for every single roll, we can use pickle to load the queryset into memory, then just use the Django queryset API for each trainee

ksungreborn:
Btw just a warning - the code is a monster...
Also, I didn't check the edge case if a trainee has multiple leave slips for the same roll...should such a thing even be allowed? =/

yipingbenjaminliu:
@ksungreborn , let's walk through this code tomorrow, I think we can drop the runtime down below 5 mins for everything

yipingbenjaminliu:
@ruthnan , @priscilagonzalez could you start testing this?
Ensure that the calculated values are accurate. If there's need for clarification, ask @ksungreborn

ksungreborn:
Looks good according to brother Andrew Li's specs, can we approve for brother Andrew? @yipingbenjaminliu.

ksungreborn:
Fixed ordering. Reports looks ready to merge. @yipingbenjaminliu @billyeh

yipingbenjaminliu:
@ksungreborn, did brother Andrew give a deadline for the report?

ksungreborn:
@yipingbenjaminliu He said asap, latest before memorial day conference, although he said it is preferable for it to be done well before then.

billyeh:
Does it still take 45 minutes??? 🤣 😆  there's no way it's going to run on prod. There's a timeout after a minute or 2.

ruthnan:
Could we generate it ahead of time and just fetch it when it needs to be sent out?

ap/reports/views.py
```diff
@@ -0,0 +1,262 @@
...
+    #print str("total roll in a week: " + str(total_rolls_in_a_week.count()))
+
+    total_rolls_in_report_for_one_trainee = int(float(number_of_days_covered) / 7 * total_rolls_in_a_week.count())
+
+    #Assuming PSRP and Monday Night Revival are classes too, then we should have 12 classes in a week
+    num_classes_in_report_for_one_trainee = int(float(number_of_days_covered) / 7 * 12)  
+ 
+    #print str("total rolls for a trainee: " + str(total_rolls_in_report_for_one_trainee))
+
+    filtered_trainees = Trainee.objects.filter(current_term__in=[1, 2, 3, 4])
```
yipingbenjaminliu: is there a reason for explicitly listing all 4 terms instead of using all?

ap/reports/views.py
```diff
@@ -0,0 +1,262 @@
...
+      except ZeroDivisionError:
+        message = "Division by 0 error."
+        rtn_data[trainee.full_name]["% Unex. Abs."] = "N/A"
+        #return JsonResponse({'bad': False, 'finalize': finalize, 'msg': message})
+
+    average_unexcused_absences_percentage = round(average_unexcused_absences_percentage / num_trainees, 2)
+    average_sickness_percentage = round(average_sickness_percentage / num_trainees, 2)
+    average_tardy_percentage = round(average_tardy_percentage / num_trainees, 2)
+    average_classes_missed_percentage = round(average_classes_missed_percentage / num_trainees, 2)
+
```
yipingbenjaminliu: for average numbers, doing calculation like this skews and weight in favor of the first few trainees it's done for. A more accurate way would just be to add each one of them, then at the end doing a divide by total trainee count

ap/reports/views.py
```diff
@@ -0,0 +1,262 @@
...
+        #print "trainee " + trainee.full_name + " has no locality..."
+        rtn_data[trainee.full_name]["Sending Locality"] = 'N/A'
+      rtn_data[trainee.full_name]["team"] = trainee.team.name
+      rtn_data[trainee.full_name]["ta"] = trainee.TA.full_name
+      rtn_data[trainee.full_name]["Gender"] = trainee.gender
+
+      trainee_all_rolls = qs_rolls.query.filter(trainee=trainee)
+      pickled_trainee_rolls = pickle.dumps(trainee_all_rolls)
+      pickled_trainee_query = pickle.loads(pickled_trainee_rolls)
+      qs_trainee_rolls = Roll.objects.all()
```
yipingbenjaminliu: general thing is that for trainees on self-attendance, the AMs may input rolls for them for auditing purposes. make sure you put in a check that for self-attendance trainees to only pull the rolls that they inputted themselvs

ksungreborn:
@yipingbenjaminliu I'll look at that. @billyeh Yes it takes ~45 minutes. I don't think I'll have time to do anything about it...might have to pass it on to someone else... =/

ap/reports/views.py
ksungreborn: I was just worried about FTTA alumni, perhaps they would be in something like term '5' or something? I don't know how we deal with alumni...

ap/reports/views.py
ksungreborn: It's outside the for loop

iamsitting:
I'm getting a keyerror % tardy, because non self-attendance trainees are not give this key-value pair Line 152

yipingbenjaminliu:
@iamsitting , could you dry up the code a little? There's a TODO tag that explains what needs to be done. Right now the uncommented code would produce a zip file as an attachment, and you can uncomment it so that it display the data instead of serving it as a zip file of pdfs

yipingbenjaminliu:
revert back to commit 9049d5a245e5b0ca2de14f3fc62ecf5ebea15c5f for this to work in full

yipingbenjaminliu:
also make sure to clear the footer new page in case trainees for a locality/team exceeds a single page

iamsitting:
New PR in #1234

PR #972, opened 2018-04-04
Interim intentions

iamsitting:
Needs some styling @yipingbenjaminliu

### DIFF
--- a/ap/ap/settings/base.py
+++ b/ap/ap/settings/base.py
@@ -181,6 +181,7 @@ APPS = (
     'graduation',
     'hc',
     'house_requests',
+    'interim',
     'leaveslips',
     'lifestudies',
     'meal_seating',
--- a/ap/ap/urls.py
+++ b/ap/ap/urls.py
@@ -63,6 +63,7 @@ urlpatterns = [
   url(r'^room_reservations/', include('room_reservations.urls', namespace="room_reservations")),
   url(r'^graduation/', include('graduation.urls', namespace="graduation")),
   url(r'^xb/', include('xb_application.urls', namespace="xb")),
+  url(r'^interim/', include('interim.urls', namespace="interim")),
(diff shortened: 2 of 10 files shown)


yipingbenjaminliu:
Could you make an admin page like HC survey with input for the following:
datetime
- 1st year return -
- 2nd year return

date:
- earliest arrival date
- term begin date

yipingbenjaminliu:
@joeyu23 , could you style the form a little

priscilagonzalez:
_Contact information_
1. Missing
   Last Name
   First Name
2. Fields are blank by default. Personal information should be pulled from the trainee.

_Interim itinerary_
1. When a new row is added, the new row copies the start date and the end date from the previous row. Should it default to blank or have the start date be the date after the previous end date?

**_Button Submit_**
1. Upon pressing submit, there is a **KeyError** at /interim/interim_intentions.

_Non-critical issues_
1. Not sure if this intentional, but on the admin page for interim intentions, the calendars are different from the open/close time and earliest arrival date/term begin date

iamsitting:
@priscilagonzalez Ready for review.

iamsitting:
- [ ] For trainee view: Datepicker doesn't show after adding a row
- [ ] For trainee view: Datetime value doesn't render consistently. For input: m/d/Y, output: Y-m-d

yipingbenjaminliu:
@iamsitting above two points addressed.

iamsitting:
@yipingbenjaminliu can we merge?

yipingbenjaminliu:
@iamsitting , the admin datetime may be a little unstable but we can test it on the live.

iamsitting:
@yipingbenjaminliu yeah it looks pretty good. Maybe we let the TA sisters test out the admin page and pass it off to @bdou

PR #989, opened 2018-04-06
save button was throwing a finalize message

yipingbenjaminliu:
@billyeh , current live server prompts for finalize message when trying to save progress for current week


yipingbenjaminliu:
Yep

PR #990, opened 2018-04-06
Ta pairings

iamsitting:
@ruthnan can you code review?


ap/leaveslips/views.py
```diff
@@ -17,19 +17,33 @@
 from accounts.models import TrainingAssistant, Statistics
 from attendance.views import react_attendance_context
 from aputils.utils import modify_model_status
-from aputils.trainee_utils import trainee_from_user
 from aputils.decorators import group_required
 from schedules.serializers import AttendanceEventWithDateSerializer
 
+TA_PAIRINGS = {
```
iamsitting: Maybe use Try/except block in case users get deleted/deactivated, etc?

billyeh:
What do you think about having the TA pairing be a field under the TrainingAssistant model?

That way we don't have to go back and edit code later, they can just change it on the admin, and it's a little cleaner, no `get_or_none`.

Issue #979, opened 2018-04-05
Exam module fixes


ksungreborn:
Exams to do
- [x] view exam score on trainee side after exam is closed  (done in branch exam-fixes)
- [x] training class in edit not showing right training classes (done in branch exam-fixes)
- [ ] finalize for grade -> after changing exam section/question BUG, trainee can’t see exam score after finalized/saved, and exam is finalized on ta side (done in branch exam-fixes)
- [x] save —> show exam score on grading page (not just for finalize)  (done in branch exam-fixes)
- [x] grade for paper response - need way to save/finalize and show on grading page (done in branch exam-fixes)
- [x] decimals needed for section (done in branch exam-fixes)
- [x] make number of exam questions required to answer for each section (done in branch exam-fixes)
- [x] admin instructions appear on trainee side should not be bold with caps (done in branch exam-fixes)
- [x] edit exam page —> preview test (done in branch exam-fixes)

Other less priority
1. open for makeup for specific trainees
2. generate makeup pdf - trainees that haven’t taken the test yet
3. sort functions (what has been graded, what hasn’t)

yipingbenjaminliu:
addressed via #1252

PR #997, opened 2018-04-07
Exam fixes

ksungreborn:
See https://github.com/attendanceproject/djattendance/issues/979 for what needed to be accomplished in this pull request. Once we merge, Raizel can start making exams for midterms (hopefully everything looks good).


iamsitting:
Exams to do
1. Trainee can see graded exam after finalizing exam (not sure if this is intentional).
2. "Study Roll" shows up as an option for training class
3. Looks good.
4. Looks good.
5. I enter a score for a trainee. Neither the table nor nor exam overview shows that. Maybe also make the table a data-table so the TA can just search for the trainee
6. Points field does not accept float (requires an integer)
7. Looks good. Maybe validate so value maxes out at number of questions
8. Looks good.
9. Looks good.

More problems:
1. It seems that every time I caused an error while making an exam, a new exam was being created (seen in the exam manager page). It's not super-critical because they can be deleted, but still needs to be resolved.

ksungreborn:
1. I think the purpose is that trainees can see the autograded questions right away. But let me check with Raizel
2. Fixed
5. It looks good on my branch. Made this table a data-table.
6. Fixed

More problems:
1. Yes, I will take a look at this later. I agree it is important

ksungreborn:
Added two fixes:
1. Filter exams shown for 2nd and 1st year trainees
2. Do not show graded exam after trainee finalizes exam

billyeh:
http://ap.ftta.lan/admin/classes/class/

ruthnan:
@billyeh, so leave it on the "Current" tab for now, and move the other apps under there (like Grad and Intentions forms) later?

billyeh:
That's a UI/TA question 😁

PR #1014, opened 2018-04-11
Leaveslips trainee filter

iamsitting:
1. Filters leave slips by trainee
2. TA Sister Approved leave slips show up with pending leave slips


ruthnan:
It works for me, I added a few style changes, but when you make the screen narrow, the dropdown options hop down to the next line before it reaches the edge of the window. It looks kinda tacky.

PR #1025, opened 2018-04-13
Datatables

iamsitting:
Using server-side datatables to display a list of Rolls, Leave slips, Events, and Schedules


yipingbenjaminliu:
okay, this datables things is pretty bomb.

yipingbenjaminliu:
@davidhtien , that's probs true, but for now I just need to set it up for the AMs so they can crunch the data and all future occurences of data. Shall we put that down for a task during service week?

billyeh:
Okay. Here we go, this is what we need. https://github.com/izimobil/django-rest-framework-datatables

yipingbenjaminliu:
@davidhtien , how far are you on DRF?

PR #1026, opened 2018-04-13
Data scripts

yipingbenjaminliu:


yipingbenjaminliu:
Check for illegitimate duplicate rolls. For every roll, check if there's a duplicate. only self attendance people can have duplicate rolls and also that the submitted_by is different.

PR #1034, opened 2018-04-14
designated services fixes

billyeh:


billyeh:
@ryanli633 feels good about it.

PR #1050, opened 2018-04-15
Attendance new look

finben:
-Changes our leave slip display to use colored vertical bars instead of the icons
-Makes it so that sister approved still displays as pending to the trainee
-Reorders fellowship slips to be above pending in priority order if there are more than one slip for an event

I have tested on firefox and chrome but may need to still test on safari


libraries/react/scss/ap_react.scss
```diff
@@ -295,30 +302,131 @@ textarea {
...
   }
 }
 
-.excused {
+.approved {
   background-color: $blue-light;
 
-  &-selected {
+  &.tardy {
+    border-right: 7px solid $yellow;
```
jontey: There is a lot of repeated border declarations. Perhaps we could use mixins in this case.

Something like this

@mixin event_border($color, $width: 7px){
  border-right: $width solid $color;
  border-left: $width solid $color;
}

and then 

&.tardy {
  @include event_border($yellow);
  ...
}

Then also the same thing for the repeating-linear-gradient

Source: https://sass-lang.com/documentation/file.SASS_REFERENCE.html#defining_a_mixin

finben:
Okay, mixins added on my latest commit

libraries/react/scss/ap_react.scss
```diff
@@ -295,30 +302,120 @@ textarea {
...
+    @include event_select_gradient($blue-light, $blue-light)
+  }
+}
+
+.pending {
+  @include event_border($blue-light);
+
+  &.tardy {
+    background: $yellow;
+    @include event_border($blue-light);
```
jontey: Is it necessary to declare this again when you already have it on line 360? Same with line 371

ruthnan:
btw, it works on Safari

libraries/react/scripts/components/WeekBar.js
```diff
@@ -20,17 +20,14 @@ const WeekBar = ({isFirst, firstStart, firstEnd, secondStart, secondEnd, period,
...
-                  <div className="col-xs-3">Denied&nbsp;<SlipStatusIcon status='D' /></div>
-                  <div className="col-xs-3">Fellowship&nbsp;<SlipStatusIcon status='F' /></div>
+              <div className="col-md-7">
+                <div className="legend row">
+                  <div className="tardy legend__tardy col-xs-3">Tardy</div>
+                  <div className="absent legend__absent col-xs-3">Absent</div>
+                  <div className="excused legend__excused col-xs-3">Excused</div>
+                  <div className="approved tardy legend__approved col-xs-3">Excused Tardy</div>
+                  <div className="approved absent legend__absent col-xs-3">Excused Absence</div>
+                  <div className="pending legend__tardy col-xs-3">Pending Leaveslip</div>
```
billyeh: Leaveslip -> Leave Slip

libraries/react/scripts/components/WeekBar.js
```diff
@@ -20,17 +20,14 @@ const WeekBar = ({isFirst, firstStart, firstEnd, secondStart, secondEnd, period,
...
-                  <div className="excused legend__excused col-xs-4">Excused</div>
-                </div>
-                <div className="row">
-                  <div className="col-xs-3">Pending&nbsp;<SlipStatusIcon status='P' /></div>
-                  <div className="col-xs-3">Approved&nbsp;<SlipStatusIcon status='A' /></div>
-                  <div className="col-xs-3">Denied&nbsp;<SlipStatusIcon status='D' /></div>
-                  <div className="col-xs-3">Fellowship&nbsp;<SlipStatusIcon status='F' /></div>
+              <div className="col-md-7">
+                <div className="legend row">
+                  <div className="tardy legend__tardy col-xs-3">Tardy</div>
```
billyeh: Do you think it'll look better if we change to `col-xs-4` to match what we had before, stretching all the way across?

finben: We tried doing the col-xs-4 and I thought it looked kind of strange since it would make a triple stack instead of a double. What do you think?

legend

finben: Okay, by setting .legend with margin 30% I think we are getting the desired effect. Two rows that stretch further across. There may be a better way to do it though but I couldn't figure out any other way

billyeh:
When I reviewed it it worked but now I have some approved slips displaying as pending on the calendar, see screen shot 
scrnshot
On the left bottom I have that approved service slip for 4/12, but the events in the afternoon on 4/12 are displayed as pending.

PR #1051, opened 2018-04-15
optimize ta list view for leave slips

davidhtien:


iamsitting:
How about handling search and pagination in the back end?

billyeh:
It's pretty impressive, tested it and there's like only 30 something queries on the whole 4,000 leave slips. But it's still really slow yeah pagination is really the only way to speed things up I think

davidhtien:
I agree, we should do pagination

davidhtien:
Nvm pagination is troublesome for combining two models, we should just get this merged in for now

PR #1054, opened 2018-04-16
Attempt at calcfix for absentee trainee roster

finben:
This is to fix our problem where the days reset every Tuesday because Monday generates a roster in which everyone is considered present. Therefore someone that has been out 9 days gets reset Tuesday to being out 1 day.  

I attempted to deep copy the Roster from Sunday on Monday and duplicate all of the values except the date. The idea is to stop our iteration cycle from breaking when it hits Monday since currently everyone is assumed present every Monday. For example if someone has been out a (Sunday) and then they are out again (Tuesday) we will need to assume that they were also out Monday and on Tuesday the report will say they have been out 3 days. By making this assumption we will give better data to the health team since in most cases if someone is out Sunday and Tuesday they probably were also sick Monday. 

However, the only problem I could foresee with this solution is if we use this same data to generate reports for sending localities on how many days a person is absent. If this is the case we will need to look at another solution since it will result in inaccurate data saying a trainee is out of more training days than would be true.


billyeh:
I think that code needed a rewrite. It didn't even have the intention of doing what was advertised.

billyeh:
Re: what days are considered out, it now counts unreported days as absences; i.e: take the current day and count back to the last time the trainee was reported present. 

I don't know if that's the intended behavior but that's seems implied by the + sign. Plus, it only reports up to the last 7 days. Anyway, that's a PM/user question. 😄

finben:
@billyeh Okay, so that sounds good then. As far as the intended behavior with the + sign signifying an unreported day. But the main thing is we have to go beyond 7 days. The whole idea is that if a trainee is out for a long time like 15 days for example that that trainee will stay on the health teams radar to inquire about them. I don't think there should even be a max number of days out but just keep reporting adding onto the days until there is a present reported

finben:
Let's do a plus sign if it's been unreported any time during the whole
period.

billyeh:
All right. Should do something like that now.

billyeh:
This should be good to go for testing this weekend 👍

PR #1064, opened 2018-04-19
Exam updates

yipingbenjaminliu:
fixed word count regardless of sections


ap/exams/utils.py
```diff
@@ -9,7 +9,7 @@
 
 @register.filter
 def get_essay_unique_id(section_id, forloop_counter):
```
billyeh: This is some really sus code 😂🕵️‍♂️. Why are we multiplying two ints like this?

yipingbenjaminliu: and something like this exists in aputils already

Issue #1053, opened 2018-04-16
TA Feedback 4/15/18


finben:
Priority Key
HP = High Priority
MP = Medium Priority

**Leave slips:**

- [ ] **Critical Issue:** The calendar no longer shows when pulling up a detailed leave slip view. @billyeh @yipingbenjaminliu 

- [ ] **HP** TA's are unable to change the events on a pending leave slip this should be editable for them  (They cannot change events on group slips. Else, everything is good)

- [x] **MP** Date of last meal out or night out not showing on detailed leave slip view for TA. e59c4853ee43a2c1688337ac4ca1da9ba7e7e12e

- [ ] **MP** There is an inherent issue with how roll and leave slip entries are being shown on the detailed leave slip page. There is no way to tell visually if there is an absence or tardy "under" a leave slip. Right now any approved leave slip changes the color of the event to blue, whether there is anything marked under or not. Please follow the principle of what we had in the old attendance server 

- [ ] **MP** When TA's process a leave slip from the detailed view it moves them on to the next leave slip in their queue which is good. However, when they process the last leave slip left in their queue it has a pop up saying saved and just refreshes them on the same page for that leave slip. Instead they should be redirected back out to the summary page and no popup is needed. 

- [ ] **MP** Numerous night out leave slips being reported as disappearing or moving to the following week on its own. Example: Connor Robinson, Samuel Gutierrez, Nestor Zepeda.

- [ ] Consider possibility of including TA private comments on list view.

- [ ] Add attendance period column on list view.
 
- [ ] TA leave slip detailed view on the detailed view, change fields to Trainee, Reason, Description, TA Comments, TA Informed, TA Assigned, and Private TA Comments.

- [x] On the detailed view, if a trainee says that he did not inform the office, the text only says "No", as opposed to "Training office informed? Yes". Example: http://10.0.8.35/leaveslips/individual/update/1646 Been fixed for a while now 3fb308eeb3bac491ee98671b1d45b0589b16c156

- [ ] Meal cleanup leave slips also excusing trainees from the meals themselves. Not an issue with the start time of the leave slip.

- [ ] Allow trainees to access previous Leave slips. Maybe a ListView for trainees.

- [ ] Saving a leave slip on the _detailed_ view does not progress to the next pending leave slip.

- [ ] When processing leave slips via the detailed view, it would be good to have it automatically ordered so that we process the leave slips dated the earliest first (not the date submitted, but the date the leave slip is for). We had this on the old attendance server.

[...]

PR #1065, opened 2018-04-19
fix permissions for saturday tas

billyeh:
Address 

> **MP** Saturday helper brothers cannot approve web access requests getting a forbidden error. May affect other modules

item on #1053. Put all Saturday and regular TAs in `training_assistant` group. Put Saturday TAs in `saturday_training_assistant` group. Put regular TAs in `regular_training_assistant` group. Use groups appropriately.

Reason for doing it this way: Saturday TAs and regular TAs should have the same permissions. Right now everything in our code uses the `training_assistant` group. When we added Saturday TAs to `saturday_training_assistant` and removed them from `training_assistant` they lost privileges. This way we just have to be specific which TAs we want.

We'll have to add everyone to the right groups after merging

(populate.py files updated only for showing how these groups are used, we wouldn't actually need to use them)

### DIFF
--- a/ap/accounts/management/commands/populate_saturday_tas.py
+++ b/ap/accounts/management/commands/populate_saturday_tas.py
@@ -20,7 +20,7 @@ class Command(BaseCommand):
         ta_obj = User.objects.get(firstname=firstname, lastname=lastname)
         print 'added ', ta_obj
         sta_perm.user_set.add(ta_obj)
-        ta_perm.user_set.remove(ta_obj)
+        ta_perm.user_set.add(ta_obj)
       except ObjectDoesNotExist:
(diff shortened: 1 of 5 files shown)


yipingbenjaminliu:
@billyeh ,are there any functional differences on the server between a Saturday TA and normal TA that would require a new group? If it's only display differences for leaveslips and what not, I'm thinking it may be more suitable to just add a new boolean field for the TA class in accounts/model. Thoughts?

billyeh:
We can't add a field under the `TrainingAssistant` model though because it's a proxy model 😆 I've tried that before.

yipingbenjaminliu:
ugh. aight, let's make a group and do this

PR #1066, opened 2018-04-19
Trainee view group slip improvements

billyeh:
* Adds trainee group select button for group slips, would be really good for trainees to have for submitting group slips e.g. team monitors, service overseers. Currently it's only available to AMs on the admin.
* Only allow submitter (and AMs) to delete group leave slips
* Move delete button and add duplicate button to slip detail view from summary


billyeh:
Shouldn't be too bad, just moving things around. 

Next steps I think for React is to look into testing 👍

billyeh:
🎉

PR #1098, opened 2018-04-27
Exam updates

ksungreborn:
Allow for makeups


ksungreborn:
Fixing unicode issue on other pages, per what Benji did on exam.html

ap/exams/templates/exams/exam_form.html
```diff
@@ -86,6 +86,8 @@
 
 
 {% block content %}
+<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
```
billyeh: Maybe we can just add this to `base.html`? https://github.com/attendanceproject/djattendance/blob/57199e25921231b49d5988d8d054fb5d9d92dc42/ap/templates/base.html#L11 Dunno why it doesn't work already with those lines.

billyeh: Actually, I don't think adding these lines helps. From base.html it knows it's UTF, the problem is the data. The symbol didn't get uploaded to the server correctly from the browser used in creating the exam

ksungreborn:
Removed the utf meta tag for now, un-reproducable on my local machine. Can we merge the changes I've made for now? It's mainly front-end exam report changes

PR #1114, opened 2018-05-03
Lord's day breakfast cleanup gs is delayed to not include breakfast

yipingbenjaminliu:


billyeh:
I suggest changing the start time of the service on the admin instead of through code like this 😆 http://ap.ftta.lan/admin/services/service/114/change/?_changelist_filters=category__id__exact%3D35%26weekday__exact%3D6

yipingbenjaminliu:
closed.

PR #1124, opened 2018-05-05
Ta edit pending leaveslip

joeyu23:
@billyeh Can you check out this PR? It adds shading for selected events for the TA view, and adds generic calendar for them to see pending group leave slips.


billyeh:
👍 !

Issue #1104, opened 2018-04-29
TA sister approved shouldn't be a leave slip status


billyeh:
It's more like an informational tag, e.g. boolean field. It's just pending, transferred, and it should retain the fact that it was approved by a TA sister. That's the issue with it being a status, since after being approved/denied by the brother they wouldn't know anymore that it had been TA sister approved in the past.

JoannaRumbley:
Also, from the Monday meeting with Jerome:

- [ ] On the detailed view, have different buttons to "Save," "Transfer," and "Approve & Transfer":
- "Save" should save and keep it in their queue 
- "Transfer" should transfer to a TA brother without adding the "TA Sister Approved" tag
- "Approve" button should always transfer to the selected TA

- [ ] On the TA brother side, the "Approve" button should always **save and approve**.

AngeloBelUCSD:
Tested:

On TA Sister Side
- Save
- Transfer
- Approve

On TA Brother Side
- **Approve** should always save and approve

Concerns:
There are two buttons on TA Sister Detail that have the same functionality: "Approve & Transfer" and "Approve".

JoannaRumbley:
Now there's only "Approve & Transfer" and no "Approve" button, so it's not repetitive. It's good. 👍

PR #1125, opened 2018-05-05
added ta-sister-approved boolean to leaveslips

bdou:
addressing #1104


davidhtien:
don't mean to barge in, but maybe we could add support for bulk sister approve here too?

PR #1127, opened 2018-05-07
fix 4th term files showing up

billyeh:
woops


yipingbenjaminliu:
4th term fellowship still comes up

PR #1138, opened 2018-05-09
try... except...

yipingbenjaminliu:
@iamsitting , could you show the object ID of the problem item instead of a None?
@billyeh, what do you think about this?


iamsitting:
Ok. It's probably better to also print out the error itself.

PR #1156, opened 2018-05-10
default datetime adjustment

yipingbenjaminliu:
remove auto adjustment by locale and force date-time format according to specified
this would set the default of the new server to view and input datetime according to the following format:
April 14, 2008 4:50pm would be shown as 04/14/2008 4:50 pm


yipingbenjaminliu:
@iamsitting , @billyeh , can we merge this?
Making the datetime input and picker consistent across the site depends on this.

PR #1168, opened 2018-05-11
Trainee can see past requests

iamsitting:
- Trainees can see past requests (Cannot edit if Denied/Approved)
- Trainees can see TA comments in table and detail view


yipingbenjaminliu:
@finben tested it, code checks out okay.

PR #1170, opened 2018-05-12
total and ta counts

iamsitting:


yipingbenjaminliu:
LG, @finben tested it.

PR #1174, opened 2018-05-12
Allows uploading for special proj

finben:
To address the problem of 

Give certain trainees the ability to upload the class files, e.g., ministry meeting verses for PSRP, etc. These would be the sisters with the "Special Projects (Outlines)" designated service. Natalie Lau, Shirleen Fang, Stacy Castillo, Deborah Chen

My thought is since we don't specifically have a "Special Projects (Outlines)" class on the server we just add these sisters to special_projects and give permission to edit class files to this special project group

### DIFF
--- a/ap/classes/models.py
+++ b/ap/classes/models.py
@@ -28,9 +28,16 @@ CLASS_CHOICES = (
     ('German', 'German'),
     ('Character', 'Character'),
     ('PSRP', 'PSRP'),
-    ('Presentations', 'Presentations'),
     ('4th Term', 'T4'),
 )
+
+# everyone has permissions for these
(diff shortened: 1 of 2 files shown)


billyeh:
Carlos did some funny stuff to allow gospel trip presentation upload. Let's try to stick to that so that special projects aren't just able to upload anything.

finben:
Okay we just need to keep in mind that apparently these special project outline sisters apparently do more than just PSRP outlines. Jerome mentioned they do need to be able to do all sorts of class files.

billyeh:
Ok. I see, thought they just needed access to PSRP file upload. In that case let's give them TA permissions 👍

PR #1175, opened 2018-05-12
Fix attendance page to support older browsers

### LINKED ISSUE #1031: Some functions don't work on Safari
1. lookup field set_all doesn't work when setting an entire team's roll for an event. 

1. Also, the dropdown to change the campus doesn't work on Safari.

jontey:
Kinda fixes #1031


billyeh:
🎉 This would be great.

I don't have an old browser to test, though 🤔

PR #1182, opened 2018-05-15
be able to unclick box

yipingbenjaminliu:
@ruthnan @priscilagonzalez @JoannaRumbley @finben 
this should allow trainees to unclick a box.


billyeh:
Do these lines not already work? https://github.com/attendanceproject/djattendance/blob/5112c116844916160830dcb1efde883bb430b2ed/ap/bible_tracker/templates/bible_tracker/index.html#L292

If not, can we get rid of one, or if we need both because of different browser versions (?) just add them both closer to each other? I can un-click boxes on Chrome on my computer already on prod.

yipingbenjaminliu:
huh, I thought the checked attribute was on label and not on the input... maybe I'm wrong.

billyeh:
Sorry I didn't mark that issue I guess I didn't see it :)

yipingbenjaminliu:
@billyeh, if it works on prod then it's good. Nevermind me. :P

billyeh:
Ok. Yeah my bad I think I randomly fixed it a while ago and forgot to mark it.

JoannaRumbley:
works on the live

PR #1183, opened 2018-05-15
Service hours field validation

iamsitting:
@yipingbenjaminliu test. review. merge

### DIFF
--- a/ap/services/views.py
+++ b/ap/services/views.py
@@ -1,64 +1,40 @@
-from datetime import datetime, date
-from dateutil import parser
-from sets import Set
-from collections import OrderedDict, defaultdict
 import random
+from collections import OrderedDict, defaultdict
+from datetime import date, datetime
 
-from django.db.models import Q
-from django.views.generic import TemplateView
-from django.views.generic.edit import UpdateView, CreateView, FormView
-from django.template.defaulttags import register
-from django.shortcuts import render, redirect
-from django.http import HttpResponseRedirect
-from django.core.urlresolvers import reverse_lazy
-from django.core.exceptions import ObjectDoesNotExist
-from django.db.models import Count, F
-from django.contrib import messages
-from braces.views import GroupRequiredMixin
-
-from rest_framework_bulk import (
-    BulkModelViewSet,
-)
-
-from rest_framework.renderers import JSONRenderer
-
-from .serializers import (
-    UpdateWorkerSerializer,
-    ServiceSlotWorkloadSerializer,
-    ServiceActiveSerializer,
-    WorkerIDSerializer,
-    WorkerAssignmentSerializer,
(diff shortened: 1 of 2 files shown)


yipingbenjaminliu:
There's also been a multiple objects returned error, take a look at that?

iamsitting:
I can't reproduce the error with those accounts. I also tried them on Safari. It seems like a rare error, so I just added a try/exception block.

yipingbenjaminliu:
@billyeh , thoughts on moving non-view functions to utils?

iamsitting:
The only one I wasn't sure about was the merge_assigns filter.

billyeh:
The multiple objects exception seems to happen with any of these models where we have a get_or_create on like form submit, e.g., designated services, absent trainee roster(https://mail.google.com/mail/u/3/#inbox/16378cf8d1f165bb). 

I think it's a concurrency issue where the object doesn't exist when A loads the page or B loads the page, then B submits, and then A submits, and gets the error because it was created in the meantime by B's submission. This is a very pervasive issue in our code 😆

iamsitting:
I took the recommendation here: https://stackoverflow.com/questions/17960593/multipleobjectsreturned-with-get-or-create, but if we do this, we will have to clear the duplicates before we migrate.

ap/services/views.py
```diff
@@ -356,7 +356,14 @@ def get_object(self, queryset=None):
     self.service = Service.objects.get(id=self.service_id)
 
     # get the existing object or created a new one
-    service_attendance, created = ServiceAttendance.objects.get_or_create(worker=worker, term=term, week=self.week, designated_service=self.service)
+    q = Q(worker=worker, term=term, week=self.week)
```
jontey: You'll need to filter by designated_service or else you will get the multiple objects error.

ap/services/views.py
```diff
@@ -356,7 +356,14 @@ def get_object(self, queryset=None):
...
+    try:
+      service_attendance = ServiceAttendance.objects.get(q)
+    except ServiceAttendance.DoesNotExist:
+      service_attendance = ServiceAttendance.objects.create(q)
+    except ServiceAttendance.MultipleObjectsReturned:
+      service_attendance = ServiceAttendance.objects.filter(q).first()
```
jontey: Also, this will cause "missing data" if ServiceRoll references a different ServiceAttendance object.

ap/services/views.py
iamsitting: Thanks! 😄

iamsitting:
@yipingbenjaminliu ready for review.

yipingbenjaminliu:
functionally is good

yipingbenjaminliu:
@billyeh , see if it effects the services functionality. For designated service hours, functionally it's effective.

PR #1208, opened 2018-05-27
Room update and tv page

zioninzion:
1. Added a new status to room reservation ('C', 'Created') that is only used in the creation of the request to enable auto-approving when the room is created by a TA while not auto-approving the requests if it is only being updated.

2. Updated tv_page_reservations to also include all the recurring events for the term.

### DIFF
--- a/ap/room_reservations/views.py
+++ b/ap/room_reservations/views.py
@@ -65,8 +65,8 @@ class RoomReservationSubmit(CreateView):
     room_reservation = form.save(commit=False)
     user_id = self.request.user.id
     room_reservation.requester = User.objects.get(id=user_id)
-    if TrainingAssistant.objects.filter(id=user_id).exists():
-      room_reservation.status = 'A'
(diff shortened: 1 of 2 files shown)


jontey:
Hi Zion, please use two space tabs for this project

ap/room_reservations/views.py
```diff
@@ -65,8 +66,11 @@ def form_valid(self, form):
     room_reservation = form.save(commit=False)
     user_id = self.request.user.id
     room_reservation.requester = User.objects.get(id=user_id)
-    if TrainingAssistant.objects.filter(id=user_id).exists():
+    #Creating a room reservation by TA will auto-approve the reservation
+    if TrainingAssistant.objects.filter(id=user_id).exists() and room_reservation.status == 'C':
```
jontey: There is a `is_TA` function you can use. I think it is in aputils/ta_utils.py

PR #1224, opened 2018-06-02
Tv page

zioninzion:


ap/room_reservations/models.py
```diff
@@ -66,7 +67,7 @@ class RoomReservation(models.Model, RequestMixin):
   frequency = models.CharField(max_length=30, choices=RES_FREQ, default='Once')
 
   # reservation approval status
-  status = models.CharField(max_length=2, choices=RES_STATUS, default='P')
+  status = models.CharField(max_length=2, choices=RES_STATUS, default='C')
```
iamsitting: A new status shouldn't be needed. The 'Pending' status already implies it's created.

ap/room_reservations/views.py
```diff
@@ -65,8 +66,11 @@ def form_valid(self, form):
     room_reservation = form.save(commit=False)
     user_id = self.request.user.id
     room_reservation.requester = User.objects.get(id=user_id)
-    if TrainingAssistant.objects.filter(id=user_id).exists():
+    #Creating a room reservation by TA will auto-approve the reservation
+    if TrainingAssistant.objects.filter(id=user_id).exists() and room_reservation.status == 'C':
```
iamsitting: Let's use is_TA(self.request.user) instead.

This form_valid function is being inherited in RoomReservationUpdate. So this logic should not apply for Update View.

ap/room_reservations/views.py
```diff
@@ -65,8 +66,11 @@ def form_valid(self, form):
...
       room_reservation.status = 'A'
+    elif room_reservation.status == 'C':
```
iamsitting: Again no need for 'C' status. If TA: status='A', else: status='P'

ap/room_reservations/views.py
```diff
@@ -127,7 +131,15 @@ def tv_page_reservations(request):
   rooms = Room.objects.all()[offset:limit + offset]
   room_data = []
   for r in rooms:
-    reservations = RoomReservation.objects.filter(room=r, date=date.today())
+    reservations = []
+    #Include recurring events
+    it_date = date.today()
+    Term_begin = date(date.today().year,2,1) if date.today().month < 8 else date(date.today().year,8,1)
```
iamsitting: Try using: Term.current_term().monday_start

iamsitting:
Looks like the auto-approve bug is fixed. However, there's a bug on the schedule:
screen shot 2018-06-10 at 3 57 29 pm
The 12:00 pm and 12:30 pm don't show up.

PR #1226, opened 2018-06-04
to accomodate event data changes

yipingbenjaminliu:
this changes roll data entry interface for AM/TM/HC as it specifies which events they can enter roll for.


yipingbenjaminliu:
Approved by @davidhtien

PR #1227, opened 2018-06-06
Grad/speaking

yipingbenjaminliu:
To allow assignment of trainees speaking
speaking trainees input portion
produces report of their speaking


iamsitting:
It looks good to me. But one more review would be better @bdou

ap/graduation/templates/graduation/outline_report.html
```diff
@@ -65,17 +65,17 @@ <h1>{{title}}</h1>
...
     <tbody>
       {% for c in data %}
       <tr>
         <td>{{c.trainee.firstname}}</td>
         <td>{{c.trainee.lastname}}</td>
-        <td>{{c.sections}}</td>
+        <td>{{c.speaking_section}}</td>
         <td>{{c.participate}}</td>
-        <td>{{c.sentence}}</td>        
+        <td>{{c.speaking}}</td>
```
bdou: Maybe I misunderstand, but why is the outline report being changed also to basically show the same thing as speaking report? Why wouldn't it stay the same, while the new fields show up under the speaking report?

bdou: other than that everything looks fine.

yipingbenjaminliu: oh, whoops. Good catch, the outline-report shouldn't have been touched

iamsitting: Good catch. @yipingbenjaminliu

iamsitting:
Please use this: https://github.com/SublimeText/TrailingSpaces to address whitespace on template files.

PR #1232, opened 2018-06-08
Fix guest web access requests

davidhtien:
Also change one class name from underscores to hyphens (delete_request_button to delete-request-button)


ruthnan:
- [ ] no start time/date on admin page
- [ ] guests can see other people's web access requests?

PR #1234, opened 2018-06-09
Attendance report updated

ksungreborn:


ksungreborn:
@billyeh You were mentioning a 2 minute timeout on the live server? Is it possible to increase this, because brother Andrew Li is asking if he can test it himself on the live server (after we merge this).

ksungreborn:
@iamsitting Think we can increase the timeout on the live server? Otherwise we can just tell brother Andrew that we'll generate the reports for him for now?

ap/reports/views.py
```diff
@@ -0,0 +1,291 @@
...
+    start_datetime = datetime.combine(date_from, datetime.min.time())
+    end_datetime = datetime.combine(date_to, datetime.max.time())
+
+    qs_group_slips = GroupSlip.objects.filter(status__in=['A', 'S'], start__gte=start_datetime, end__lte=end_datetime)
+    t.end()
+    # filtered_trainees = filtered_trainees.filter(firstname="David")  # Test line
+
+    localities = Locality.objects.all()
+    for locality in localities:
+      if str(locality.city) == "Richmond, VA":
```
billyeh: Richmond doesn't seem to be handled fully correctly. There's only one file in the zip output, with trainees from Canada and Virginia.

ap/reports/views.py
```diff
@@ -0,0 +1,291 @@
...
+    # qs_rolls is the queryset of all pertinent rolls related to the filtered trainees in the date range
+    pickled_query = pickle.loads(pickled_rolls)
+    qs_rolls = Roll.objects.all()
+    qs_rolls.query = pickled_query
+
+    # get all group slips in report's time range with the specified trainees that have been approved, this is needed because groupslip start and end fields are datetime fields and the input is only a date field
+    start_datetime = datetime.combine(date_from, datetime.min.time())
+    end_datetime = datetime.combine(date_to, datetime.max.time())
+
+    qs_group_slips = GroupSlip.objects.filter(status__in=['A', 'S'], start__gte=start_datetime, end__lte=end_datetime)
```
billyeh: Should be slip start <= period end and slip end >= period start.

ap/reports/views.py
ksungreborn: Oops, my bad. Fixed

ap/reports/views.py
ksungreborn: I was thinking long and hard about this...isn't it the same? =/

billyeh: Yeah, it's hard to figure out. Basically the code I commented filters out leave slips that aren't completely within the period, e.g. one that begins before the period but ends within it.

yipingbenjaminliu:
some work i'm doing on this module to allow us to use it on the live server without changing nginx timeout value
- sends a list of trainee ids, locality ids, team names to front end and render empty page
- automatically for each trainee id to send ajax request to backend to independently process computation and utilize backend to frontend communication for parallel processing
- render results on frontend when backend computation is sent back
- have a download button available that would send results all as string to backend and render it into a zip file of pdfs
- eventually have options on frontend to change attendance report pull options

iamsitting:
@yipingbenjaminliu Beautiful! I just used this for import. I love it. This will buy us time until we can implement celery. We also need to fix the math by using Kevin's. Let me know when you take a pause and I can try to address that.

yipingbenjaminliu:
@iamsitting , I'll clean up this module and implement it till it's fully workable including cleaning up the code for both frontend and backend. I probably won't finish before term starts.

iamsitting:
@yipingbenjaminliu I'm gonna test these reports with what we sent out to Andrew at the end of last term to verify the math and lint the code while I'm at it.

iamsitting:
@yipingbenjaminliu Functionally, this is very good. I love progBar!
- I'm not sure if this is still a WIP, but the numbers aren't right.  I didn't see why we deleted Kevin's code in ffa2ee267e4043ab94e6b027913e861c5e733327
  - Were you planning to refactor Kevin's math/queries from: https://github.com/attendanceproject/djattendance/blob/c414ebd0ea54bf737de09ffdfcb431aca8c19c0f/ap/reports/views.py#L49
- We also want to address the codacy issues for the JS file. This is minor, but we still want some kind of code standard.

yipingbenjaminliu:
@iamsitting ,how are you comparing the numbers? The numbers should be off here and there because of changes to the events type and event monitor and how the code is doing the calculations. I'll put in some more comments in the code to clarify how all the calculations are done.

I deleted Kevin's code cause I refactored it into the existing parts, so if it's deleted it means that it should exist somewhere in the current file, I used most his logic and most parts of his code, I could incorporate all the try-except as well just for robustness.

I'll eventually produce a front-end that's "exactly" like the old server with the options to customize the headers for the produced file (eg: unexcused absences by type, lifestudies, etc...), but for the time being, this seems to be what the TAs are comfortable with.

I also need to change how the ajax queries are being fired from the frontend. If their all fired at the same time, some of the ones dangling at the end may get a time-out error through nginex.

iamsitting:
@yipingbenjaminliu I run your code on the whole term (I got the sql dump before I left). Then I compare your numbers with Kevin's (the report which was sent to Andrew, you were CC'ed). I'm assuming it's more ore less the same dataset. 

[testdata.xlsx](https://github.com/attendanceproject/djattendance/files/2263371/testdata.xlsx)

These are my team's results from your and Kevin's math. It's definitely possible I'm doing something wrong...

yipingbenjaminliu:
@iamsitting , yea, the numbers are definitely different from what was sent to Andrews because of the change in events calculation. I'll include the comments and write up a small document for it to explain how the workflow goes as it's starting to get really massive.

I've pretty much hit a checkpoint with this branch, the rest can be a separate branch, including the stuff for the frontend

yipingbenjaminliu:
@iamsitting , i'm not sure if this branch with the current state will be able to run without timeout errors because it throws all the ajax requests at once and wait for the server to responds. Functionally, it works and we can try it on the live to see if it works. If not, there's a very easy fix that can be done on the front end that limits the number of live ajax requests we have

iamsitting:
@yipingbenjaminliu 
> the change in events calculation.

What is that? So which numbers are right? To me Kevin's report is more correct because of the "missed classes" column alone (yours show a bunch of zeros). But his numbers also look right (I can at least vouch for my team's data).

> I've pretty much hit a checkpoint with this branch

So you're done making changes to this branch?

> we can try it on the live to see if it works

Yes, I can merge it into live and try it out and let you know. But I'd like to settle the math issue first.

yipingbenjaminliu:
@iamsitting 
so the way I've written the code right now for classess missed is that it excludes any classes with a valid leaveslip, I believe Kevin's calculation was just all classess missed regardless of leaveslips.

I'm done making changes to the branch

iamsitting:
@yipingbenjaminliu 
This is from Andrew's email referring to Fall 2017:
>Last term, the averages were:
   -- An average of 0.18 % of Unexcused Absences were recorded for trainees over the term.
   -- An average of 2.38 % of Sickness Absences were recorded for trainees over the term.
   -- An average of 1.39 % of Rolls Tardy were recorded for trainees over the term.
   -- An average of 6.25 % of Classes Missed were recorded for trainees over the term.

Now Kevin's stats for last term:
>• Average % Unex. Abs.: 0.17%
• Average % Sickness: 1.97%
• Average % Tardy: 2.0%
• Average % Classes Missed: 12.73%

This is what I get with the latest pull:
>unexcused_absences_percentage: 0.20%
    sickness_percentage: 2.80%
    tardy_percentage: 0.16%
    classes_missed_percentage: 0.44%

Your tardy and CM % are way too low.

What do you think about calculating those without the leave slips? I could do it. (I could also confirm with Andrew.)
________
________
One more request: Could you attach the averages to each page of the report? I think you could do it much faster than I could.

yipingbenjaminliu:
Okay, got it. I just got to the SST location, so I'll probably be absent until Saturday afternoon. Sorries. :P

iamsitting:
Nope, not urgent. Enjoy bro.

PR #1235, opened 2018-06-09
Exam fixes

francestsenn:
- Matching questions: For Matching questions, we want the ability to have more answer choices than questions (i.e. 9 matching questions and 10 possible answers to select from). The current supposed fix is to add an answer, but leave the point and prompt fields blank. The problem may be on line 234 in utils.py in the exams folder. Also look into hstore for bringing things from the back end to the front end.
- Keep everything on 'Exams' tab and drop 'Current' Tab
- In exam overview, change the Average score to display only 2 decimal values.
- Raizel's feedback on formatting (email from June 5):

> 1) I'm still not sure if what the trainees see is what I see when I preview exam. For the instructions for CYPC, I would like the instructions to appear as a list, like this:
>
> As a reminder, the topics related to the campus work are as follows:
> 1.   The Full Gospel according to the Heavenly Vision
> 2.   The Two Main Aspects of the Lord’s Recovery Today
> 3.   Preaching the Gospel on the College Campuses by Prayer
> 4.   Preaching the Gospel on the College Campuses by the Spirit and the Word
> 5.   Preaching the Gospel on the College Campuses by the Homes
> 6.   Contacting People by Demonstrating the Spirit
> 7.   Fruit-bearing: The Preaching of the Gospel on the Highest Plane
> 8.   Kernel of God’s Gospel—the Indwelling Christ
> 9.   A Proper Daily Living and Universal Love
> 
> 2) Small formatting points: would it be possible to make "Instructions:" bold so that it stands out more. "Points" does not need to be bold.
> For answers, can it be formatted so that it appears closer to the question it's answering, rather than to the next question?


yipingbenjaminliu:
average score one decimal out of two decimal place, can we make it consistent?
image

yipingbenjaminliu:
i don't think the matching questions is fully functional?
image

yipingbenjaminliu:
for the grading part,
image

yipingbenjaminliu:
also, filtered class, when returned to does not filter according to selection
image

yipingbenjaminliu:
also, what's Raizel's feedback on formatting

yipingbenjaminliu:
required number of questions to answer can be greater than question count, please validate that field
image

PR #1245, opened 2018-06-11
doesn't send absentee roster on monday

yipingbenjaminliu:
modification on cron job so it still runs backup on monday but doesn't send absentee trainee roster


billyeh:
Has it been sending on Monday? We have this code https://github.com/attendanceproject/djattendance/blob/dev/ap/absent_trainee_roster/utils.py#L106-L107 so maybe we can decide to keep one or the other.

yipingbenjaminliu:
@billyeh , yea, I realized that those lines actually still results in cron_job calling absentee roster trainer every monday and those data is being produced, just not sent. So every monday, we're making useless records. We can probably remove those lines and just use the ones above where cron_job tasks are modified instead.

PR #1247, opened 2018-06-13
Better get_attendance_record and panels calc fix

iamsitting:
Fixes get_attendance_record to match front end counts.
Panels should now display matching personal attendance counts.


iamsitting:
@samvictor232 please review... thanks.

PR #1256, opened 2018-06-14
Lockout ss

ruthnan:
hot fix for not allowing service schedulers to edit permission groups

I'm not sure why the TAs were removed twice (aside from if the AM permission group was 'saved' twice), but we need to give them access to change rolls that's not tied to being an attendance monitor.


yipingbenjaminliu:
um, you're bringing in commits from another branch. clear out those commits please

ruthnan:
sorry, @yipingbenjaminliu - here you go.

billyeh:
It was probably just saved again after we fixed it the first time. Let's merge this to prevent it from happening again at least.

PR #1265, opened 2018-06-19
return queryset rather than list from get_workers

billyeh:
Fixes a bug with service scheduling where designated services call `get_workers.count()` on an empty list for non-active services.


ryanli633:
I think this might be what is causing the "Make Schedule" button to throw a 500 error (Type Error). So, the "Make Schedule" button doesn't work right now. The service schedulers brought this to my attention. Not sure if we should merge this PR in ASAP to address this problem.

PR #1267, opened 2018-06-19
allow self-attendance for semi-annual study

yipingbenjaminliu:
tried to mimic the old server, it's date based, so when testing it, try to move local time into the semi-annual week to test it


iamsitting:
Can we make this it's own module or at least its own view? I think it's better to keep this separate from the attendance module.

yipingbenjaminliu:
yea, that's probably better. bro I'm studying for an interview tomorrow and on Friday, I won't have time for this until Sat.

iamsitting:
Closing for #1283

PR #1268, opened 2018-06-20
TV page clock update

davidhtien:
start using time.is

see: https://time.is/widgets

### DIFF
--- a/ap/room_reservations/static/room_reservations/js/clock.js
+++ b/ap/room_reservations/static/room_reservations/js/clock.js
@@ -47,6 +47,7 @@ function time_is_widget() {
             seconds: "%s",
             AMPM: "%A"
         };
+        // defaults
         var template = "TIME";
         var timeFormat = "%H:%i:%s";
         var dateFormat = "%Y-%d-%m";
@@ -104,11 +105,9 @@ function time_is_widget() {
     };
 
     this.cb = function(t, r, a) {
-        console.log('cb');
         var rpT = new Date()
         var n = 0;
         time_is_widget.timeDiff = rpT.getTime() - t - Math.round((rpT - r) / 2);
-        console.log("cb config", this.config);
         for (var i in this.config) {
             this.config[i]["vector"] = a[n];
             n++;
@@ -154,7 +153,7 @@ function time_is_widget() {
                     s: c["vector"][5]
                 }
                   , h = {};
-                for (j in g) {
+                for (var j in g) {
                     h[j] = addLeadingZero(g[j]);
                     h[j + "H"] = h[j];
                     h[j + "M"] = "AM";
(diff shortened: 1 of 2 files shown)


davidhtien:
clock.js is pretty unreadable :/ unfortunately but it does work, I would refactor but it might not be worth it, also this adds a dependency on time.is which is another thing to consider, I don't think they'll be going anywhere soon but you never know

ap/room_reservations/templates/room_reservations/tv_page.html
```diff
@@ -7,254 +7,232 @@
   <link rel="stylesheet" href="{% static 'room_reservations/css/font.css' %}" />
   <link rel="stylesheet" href="{% static 'room_reservations/css/weather-icons.min.css' %}" />
   <script src="{% static 'room_reservations/js/weather.js' %}"></script>
-</head>
```
davidhtien: I just did some reformatting here, placing everything in the head tag and indenting

ap/room_reservations/templates/room_reservations/tv_page.html
```diff
@@ -7,254 +7,232 @@
...
+    hours = hours % 12;
+    hours = hours ? hours : 12;
+    minutes = minutes < 10 ? '0' + minutes : minutes;
+    var strTime = hours + ':' + minutes + ' ' + ampm;
+    return strTime;
   }
-  $('#daft').text(forecasts[2].day);
-}
 
-function updateClock()
```
davidhtien: I deleted this function

ap/room_reservations/templates/room_reservations/tv_page.html
```diff
@@ -266,12 +244,7 @@
   <div class="row">
 
     <div class="col-sm-7 text-center">
-      <div id="clock-container" class="row">
-        <div class="col-sm-12"><span id="clock"></span><span id="ampm"></span></div>
-      </div>
-      <div id="date-container" class="row">
-        <div class="col-sm-12"><span id="date"></span></div>
-      </div>
+      <span id="Anaheim_z14e"></span>
```
davidhtien: this is where clock.js inserts

ap/room_reservations/static/room_reservations/js/clock.js
```diff
@@ -0,0 +1,213 @@
...
+                }
+                if (W == 0) {
+                    W = 52;
+                    if (new Date(y - 1,0,1).getDay() == 4 || new Date(y - 1,11,31).getDay() == 4) {
+                        W = 53;
+                    }
+                }
+                if (p["W"] == "hy") {
+                    if (W == 1) {
+                        pw = "1-ին շաբաթ";
```
iamsitting: What is this?

davidhtien: bruh I wish I knew

davidhtien: ok from a little googling it seems like this has to do with the Armenian calendar... I'll take it out I don't know why it was included in their widget script

davidhtien:
Alright @iamsitting can you take another look? I want to merge it in and test it out to see if it's accurate over multiple days, we can check in again on saturday during service time

davidhtien:
Actually let's try #1273 first, don't merge this yet

PR #1273, opened 2018-06-21
Make tv page clock more accurate

davidhtien:
change from setInterval to setTimeout

this method gets rid of the gradual drifting due to extra milliseconds eventually stacking up


billyeh:
The thing is it won't work until we are using the new server for the display page I think @zioninzion was working on that?

Issue #1272, opened 2018-06-21
Audio - week 0 Gateway timeout


JoannaRumbley:
Trainees cannot see/access audio files for Week 0 due to a 504 gateway timeout error.

billyeh:
@JoannaRumbley #1297 ended up fixing this for good for now

PR #1275, opened 2018-06-23
Audio naming

### LINKED ISSUE #1186: Naming the Audio
Continuing from discussion on #1133 

Need to iron out the naming format and make sure each class is labeled correctly (currently, we label them with week #, class name, and speaker). 
- Week 5 God's Economy audio doesn't have the speaker name.
- Monday Revival Meetings are by "Selection," "Reports," or "Presentations." Should these all use the same word?

(This is a minor issue, but it does matter more from a UX perspective. We want people to feel the server is clear and consistent.)

billyeh:
This helps with #1272 and fixes #1186, introducing tests to make sure there's no regression in terms of parsing the audio names. Uses `pytest-django`. We have some integration tests (which currently don't work) but I don't know how much longer we can go without adding unit tests. Let's start.

### DIFF
--- a/ap/absent_trainee_roster/utils.py
+++ b/ap/absent_trainee_roster/utils.py
@@ -103,8 +103,6 @@ def calculate_trainee_absent_freq(date):
 
 def send_absentee_report(year, month, day):
   d = date(int(year), int(month), int(day))
-  if d.weekday() == 0:
-    return "Today's a monday, so no absent trainee roster"
   ctx = build_report_ctx(d)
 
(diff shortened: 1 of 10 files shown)


iamsitting:
Just some rendering issues I found. Some aren't critical though.
screen shot 2018-06-23 at 10 41 38 am
screen shot 2018-06-23 at 10 42 06 am
screen shot 2018-06-23 at 10 43 36 am

billyeh:
Ok. I fixed the rendering problem with a Morning Revival Fellowship class.

PR #1279, opened 2018-06-23
Fix finalize self attendance

joeyu23:
@billyeh Can you review this? Should fix the issue where trainees, after finalizing attendance, see the attendance record inputted by the attendance monitors rather than their own attendance. 
@ruthnan @priscilagonzalez Can you all test this fix?


billyeh:
Accidentally had some other files in these commits so I guess when I removed them this got closed. Moved to another PR :) #1280

PR #1280, opened 2018-06-23
fix finalization show correct rolls in response

billyeh:
@billyeh Can you review this? Should fix the issue where trainees, after finalizing attendance, see the attendance record inputted by the attendance monitors rather than their own attendance.
@ruthnan @priscilagonzalez Can you all test this fix?

from #1279


yipingbenjaminliu:
move to #1281

yipingbenjaminliu:
temp fix

PR #1286, opened 2018-06-27
Seating chart roll upgrades

davidhtien:


yipingbenjaminliu:
small problem, seems like I can't get past absent...

PR #1287, opened 2018-06-27
exam - enforce data constraint to prevent duplicate entries

yipingbenjaminliu:
DON'T MERGE until data is cleared up


yipingbenjaminliu:
@iamsitting , @francestsenn , consider merging this?

PR #1289, opened 2018-06-27
allow training assistants AM access on react

billyeh:
Related to #958, then #1256. Once we removed people from groups automatically, TAs began to lose attendance monitor group. They shouldn't need it so this PR allows that.

Also just a lot better than relying on a group ID being 4 all the time.


billyeh:
@ruthnan I think the only problem we ran into was the attendance monitors being cleared of TAs, anything else? Did Dennis lose privileges he needed?

Issue #1282, opened 2018-06-23
Service Week Issues - TA Feedback List 6/23


finben:
To give everyone some ideas of things to do for service week. Here is an updated list of bugs from brother Jerome. This list is by no means all inclusive but it is up to date. Overall this has been a successful term and for all of your encouragement in TA bugs alone we have solved 103 bugs total so far into this term. Many big fires have been put out this term and it has been an 'exciting' shakedown cruise. As an additional main objective for this service week we will attempt to upgrade three things.  Our server version, Python to Python 3, and Django to 2.0. this service week. Happy Service Week! -Ben

**Attendance Rolls**
- [ ] Add force page when submitting attendance late.
- [ ] Need unfinalize roll button. Relatedly, the finalize button should be blanked out if attendance is finalized. @yipingbenjaminliu #1179
I also don't see an unfinalize button...is it done, but not on the live server yet?
- [ ] TA input attendance - Unclear if clicking "Finalize" does anything. (Maybe have a "Finalized" notification or make the "Finalized" button grayed out.) On the old server, the button blacks out when attendance is finalized.

**Extra Fields**
- [ ]  Add "Does not count against perfect attendance" check box and functionality. 
- [ ] Does not affect perfect attendance will be a built-in leave slip category” (Automatic for Conferences, team trips, house meals, service, fellowship, and allow for TA override)

**HC Survey List**
- [ ] List should be organized among brother's and sister's houses. Currently no order to the list

**Interim Intensions**
- [ ] On the interim intentions report, for some trainees the "Intention" reads a mix of blanks, "None," and "Serve Full-time: College Campus in the U.S." "Serve Full-time" is not an option for trainees to select under "Returning" on the form. It is only happening to some, not all, the returning trainees (see Adolfo Fernandez, Alejandra Ferrel, and Alex Cantu). The "Intention" for all returning trainees should all read "None."
- [ ] The form on the old attendance server asked for specific information from the trainees based on which option they chose under Post Training Intentions (which translates into "Intentions" on the report in the new server), and not just "Explain:". Most trainees did not provide information such as the locality they will end up in for work, what school and program they will be enrolled in, what campuses they will be serving at, or when they will start serving.

**Leave Slips - Group Slips**
- [ ] TA's are unable to change the events on a pending leave slip this should be editable for them (They cannot change events on group slips. Else, everything is good) @joeyu23
[...]

PR #1293, opened 2018-06-27
Maintenance 1154

priscilagonzalez:
#1282 Other items: changing per TAs' requests


yipingbenjaminliu:
@priscilagonzalez, adjusting request_list/list.html would change the reports for linens, maintenance and framing. is that intentional to have the following headers for all three reports?
[Date Requested, Name, Phone, House, Room, Type, Description, Comments, Status, View]

PR #1295, opened 2018-06-28
Ta groupslip detail view fixes

davidhtien:
- [x] Make selected groupevents shaded when viewing in TA detail view
- [x] Select all events between when you click on two noncontiguous groupevents
- [x] Deselect all but clicked event when clicking on a groupevent that is already shaded in a group of three or more shaded contiguous groupevents
- [ ] Update TA detail view form fields for start and end time based on selected groupevents


iamsitting:
Group slips don't save if I edit the selected group events.

PR #1296, opened 2018-06-28
1282/room reservation notif

ruthnan:
Room Reservations:
- [x] Defaulting search for rooms on TA list to "Marked" and "Pending" by setting regex: true
- [x] Added notification for room reservation requests being marked for fellowship


yipingbenjaminliu:
@ruthnan , actually, could you move the commits to a new branch?

PR #1306, opened 2018-06-29
Schedules changes

yipingbenjaminliu:
The goal of this PR is to provide the Non-technical access for schedule operations, mainly schedule creation, update, and delete (CRUD operations). There is a difficulty that has plagued this project as schedule provide the underlying base data framework for which rolls is based on and almost all other components are based on rolls.

The test cast results should allow trainees with the set permissions to modify and adjust schedules as they see fit without any unintended results that would cause rolls or individual leaveslips to disappear. 

Implementation items:
### - [x] afternoon class change
this feature provides a page to moves trainee's afternoon class schedule. They can be moved between Greek I, Greek II, Character, German I, German II, etc... It should not delete any existing rolls with possibly attached Individual Leaveslips but should transfer it to the designated class

#### dev notes:
this is a specific case of schedule update with strict known parameters that allows rolls update
- template afternoon_class.html, simple interface with three fields, trainees to be switched to a different afternoon schedule, which class to move them to, starting from which week
- view function AfternoonClassChange:
  -  predefines event choices in get_form_kwargs
  -  queries for list of events by specific known parameters
  -  adds a stub choice for initial default
  -  form_valid outputs success or failure message
- form AfternoonClassForm, defines input fields with widget
- util method afternoon_class_transfer
  -  queries for effected rolls 
  -  check to make sure new schedule is valid and matches the criteria
  -  find an already made transfer schedule
  -  if not found, create one
  -  add transfer trainees onto new schedule
  -  move rolls over
  -  output message with names of all trainees

#### test case:
1. trainee that has rolls for afternoon class at week X, move that trainee to a different afternoon class starting from week X. Personal attendance should not reflect that move along with their rolls moved
2. trainee that has rolls with leaveslips for afternoon class at week X, move that trainee to a different afternoon class starting from week X. Personal attendance should not reflect that move along with their rolls and leaveslips
3. move trainee into afternoon class A at week Y. move another trainee into afternoon class A at week Y, both trainees should now be on the same schedule
4. move trainee into afternoon class A at week (Y+1), this trainee would then be on a different schedule

### - [x] schedule creation
this feature provides a form for the user to input the details of the schedule for it to be created. It should prompt for rolls deletion if the user is attempting to create a schedule that would override existing events that have rolls assigned to those events for those trainees on the created schedule

#### dev notes:
- template admin_form.html
  -  ajax function to make a server call to delete conflicting rolls
  -  display of rolls that needs to be deleted in order for the schedule creation to happen
  -  hide the default error view from boostrap_form using CSS, I couldn't find a way to disable it internally
- view function ScheduleAdminCreate that simply extends ScheduleCRUD and CreateView:
  -  form_invalid for ScheduleCRUD simply takes the information in form.errors that's encoded for error rolls to be parsed fed to the template for rendering, it also display which operation (CRUD) was taken
- form CreateScheduleForm that extends BaseScheduleForm:
  -  BaseScheduleForm predefines initial values for load, and associates widgets for some
  -  save method is needed for the clean method to be called
  -  Create uses a slightly different clean method compared to Update and Delete because there are no initial values.
  -  The logic is as the following:
      -  given trainee set A, look for all the schedules that have any one within trainee set A that is not a gorup schedule or has priority greater than form priority
      -  get the event information from all those schedules, call this interested_eventsList
      -  now get all events in interested_eventsList that overlaps with form events, these are the potential events that this new schedule may override
      -  query for potential rolls with these events and trainee set A, this queryset of rolls needs to be deleted before this schedule can be created

### - [x] schedule updates
this feature provides a form for the user to input the details of the schedule that needs to be adjusted. It should prompt for rolls deletion if the adjusted schedule would cause any rolls to become ghost records as the events that they were once assigned to are no longer applicable to those trainees. This may be due to the the changes in the trainee field, or the events field, or the priority field, or the weeks field.
#### dev notes:

- template admin_form.html, same as above
- view function ScheduleAdminUpdate that simply extends ScheduleCRUD and UpdateView:
  -  form_invalid same as above
  -  form_valid allows using ScheduleAdminUpdate for both update and delete
  -  prior to this, we used a separate DeleteView as a form but the rolls validation required form information  
- form UpdateScheduleForm that extends BaseScheduleForm:
  -  both Update and Delete eventually will call the util method validate_rolls_to_schedules with a list of schedules, a trainee set, weeks, and a queryset of rolls; Update and Delete differ in how these parameters are produced
  -  check if weeks have change from form initial, if so set weeks to max length, the min and max between initial and changed
  -  check if trainees have change from form initial, if so trainee set is trainees of form initial plus trainees of form change
  -  create list of schedules and exclude form initial and add form changed, this is to essential mock the system with what the schedule change would look like
- util method validate_rolls_to_schedules:
  -  this method uses the same function for querying the data for personal attendance (EventUtils.collapse_priority_event_trainee_table)
  -  produces a giant OrderedDict that represents the entirety of schedule for given trainee set
  -  for each roll, check if that trainee has that specific event on that date
  -  output the rolls where the trainee no longer has that event on that date  

### - [x] schedule deletion
this feature is a button at the end of the form for updating the schedule that allows the user to delete this schedule. It should prompt for rolls deletion if the user is attempting to delete a schedule that would cause certain rolls to no longer be accesible because the trainee no longer will have those events
#### dev notes:

- template admin_form.html, same as above
- view function ScheduleAdminUpdate, same as above:
  -  form_valid, if the delete button is pressed and the form checks out, delete and go to schedules-table
- form UpdateScheduleForm that extends BaseScheduleForm:
  -  same as above
  -  just remove the schedule and call the util method

#### test case:


yipingbenjaminliu:
@iamsitting , updated. Let me know if there's need for more clarification.

iamsitting:
Just to clarify: There are two things going on?
1. CRUD for schedules
2. Afternoon schedule changes

yipingbenjaminliu:
well, schedules CRUD is general, afternoon schedules changes are a specific case that allows rolls transfer rather than deletion

ap/schedules/utils.py
```diff
@@ -1,8 +1,102 @@
...
+  roll_ids = []
+
+  current_term = Term.objects.get(current=True)
+  w_tb = EventUtils.collapse_priority_event_trainee_table(weeks, schedules, trainee_set)
+  potential_rolls = rolls.order_by('date', 'event__start')
+  for roll in potential_rolls:
+    key = current_term.reverse_date(roll.date)
+    evs = w_tb[key]
+    if roll.event not in evs or (roll.event in evs and roll.trainee not in evs[roll.event] ):
+      roll_ids.append(r.id)
```
iamsitting: roll.id, not r.id

ap/schedules/utils.py
```diff
@@ -1,8 +1,102 @@
...
+  whole_term_sch = Schedule.objects.none()
+  potential_sch = Schedule.objects.filter(events__in=new_ev)
+  for sch in potential_sch:
+    weeks = [int(i) for i in sch.weeks.split(',')]
+    if min(weeks) == start_week:
+      new_sch = sch
+    if min(weeks) == 0:
+      whole_term_sch = sch
+
+  new_evs = whole_term_sch.events.all()
```
iamsitting: Error at L45, whole_term_sch is QuerySet

ap/schedules/utils.py
```diff
@@ -1,8 +1,102 @@
...
+  # obtain involved rolls that needs to be modified here
+  # make sure the django query is executed for both old_evs and rolls otherwise they mutute because of laziness or re-eval
+  old_evs = set(ev for ev in trainee.events if ev.class_type=='AFTN' and ev.monitor=='AM')
+  old_sch = Schedule.objects.filter(events__in=old_evs, trainees=trainee)
+  old_priority = max(list(old_sch.values_list('priority', flat=True)))
+  old_rolls_ids = list(Roll.objects.filter(trainee=trainee, event__in=old_evs, date__gte=date).values_list('id', flat=True))
+
+  # get the new event and the new schedule that the trainee will be transferred onto
+  new_ev = aftn_evs.filter(code=e_code)
+  new_sch = Schedule.objects.none()
```
iamsitting: You define new_sch as QuerySet, but then treat as a Schedule object later causing an error in L70.

yipingbenjaminliu: I need to rewrite this method...this was written a while ago

ap/schedules/utils.py
yipingbenjaminliu: I need to rewrite this method...this was written a while ago

yipingbenjaminliu:
PR isn't ready yet, still working on it

ap/schedules/forms.py
```diff
@@ -52,31 +59,127 @@ class ScheduleForm(forms.ModelForm):
...
+
+# def validate_rolls_to_schedules(schedules, trainee_set, weeks, rolls)
+
+class CreateScheduleForm(BaseScheduleForm):
+
+  def clean(self):
+    cleaned_data = self.cleaned_data
+
+    new_schedule_instance = deepcopy(self.instance)
+    new_schedule_instance.trainees = cleaned_data['trainees']
```
iamsitting: For manytomany relationship, you must save the object before adding the relationship: https://docs.djangoproject.com/en/2.1/topics/db/examples/many_to_many/
This is of course only a problem for creating schedules, not updating them.

ap/schedules/views.py
```diff
@@ -178,66 +178,66 @@ class EventAdminDelete(EventCRUDMixin, DeleteView):
...
+  def form_invalid(self, form, **kwargs):
+
+    context = self.get_context_data(form=form)
+
+    error_data = json.loads(form.errors.as_json())
+    errors_list = error_data['__all__']
+    for error in errors_list:
+      if error['code'] == 'invalidRolls':
+        rolls_to_delete = error['message']
+        roll_ids = [int(s) for s in rolls_to_delete[1:-1].split(',')]
```
iamsitting: rolls_to_delete = u'[<Roll: ID 88666 [2018-09-06] ID 1102: Thursday 3 [18:45 - 21:14] Study @ [A] Raymond Chang>,  Study @ [A] Raymond Chang>]'
I'm assuming you were expecting a csi string of ids, because your parsing expression throws an error.

ap/schedules/forms.py
yipingbenjaminliu: I'm not saving it. I'm creating a ghost schedule to evaluate the validity of the update.

ap/schedules/views.py
yipingbenjaminliu: I'm expecting an array of integers that are roll IDs expressed as a string tossed from the clean method in form, where the views takes them in form_invalid and renders them on the template

ap/schedules/forms.py
iamsitting: Yeah. That's the problem. Otherwise, I get "<Schedule: ID None: [None]  - None schedule>" needs to have a value for field "id" before this many-to-many relationship can be used.

ap/schedules/utils.py
```diff
@@ -1,8 +1,97 @@
...
+  all_names = ''
+  if not found:
+    old_sch = potential_sch.first()
+    new_sch.pk = None
+    new_sch.save()
+
+    new_sch.import_to_next_term = False
+    new_sch.name = new_sch.name + ' - transfer'
+    new_sch.comments = old_sch.comments + ' // used for transfers'
+    new_sch.priority = old_sch.priority + 1
```
iamsitting: The priority should also consider the current schedule. i.e Char to GK1
Now: new GK1 pr = old GK1 pr + 1
Should be: new GK1 pr = Char pr + 1

iamsitting: It's rare, but it happens.

yipingbenjaminliu:
Are both the test case above from afternoon schedule change or from CRUD operators?

iamsitting:
For the error message on updating schedules:
  The following rolls needs to be deleted before this schedule can be updated

Schedules are updated anyway.

yipingbenjaminliu:
@iamsitting, test cases resolved. Test again please, see if you can find anything.

iamsitting:
Test case: Absent for Remote (Tuesday) and Char (Thursday)
Transferred to Germ 1
Schedules change is made, but rolls are not preserved. (Seems to only happen to those with remote event)

iamsitting:
I would like @francestsenn to test this as well.

francestsenn:
Creation -- Does not delete or prompt for roll deletion when there is a schedule conflict.
Update -- Works for the most part. Cannot update the weeks field.
Deletion -- Looks good. I'll add a confirmation since I've already accidentally deleted a schedule... 
😬

francestsenn:
Should be ready to merge. Problems with creating the schedule were related to having events that started before and ended after the original events @ryanli633

PR #1318, opened 2018-07-02
Ticker

zioninzion:
Added 'On TV Page' option in announcements which will be shown on the TV page in Room Reservation.


iamsitting:
screen shot 2018-07-04 at 2 41 10 pm

@xuefeizheng and @isaactsou Could we get a design for the ticker? Maybe some options so we could show Jerome.

zioninzion:
Resolved an issue with not correctly reloading the page

zioninzion:
Broke up js and css codes into load-all.js and styling.css files in static folder

davidhtien:
This works, here's what it looks like:
screenshot 2018-10-03 17 53 32
Could we run this by a designer? I feel like we could do something a little more aesthetic

Issue #1323, opened 2018-08-08
/lifestudies/<pk>/detail-discipline is accessible to all. Any trainee can purge his/her life studies


alextricity25:
Steps taken to produce:
1. Created trainee user with the name test.
2. Assigned a life study to said user
3. Attempted to submit a life study summary for discipline assigned
4. Obtained the primary key of the life study record from URL
5. Accessed `/lifestudies/<pk>/detail-discipline`
6. Decreased penalty count successfully

Suggestions for remediation: User permissions should be checked upon a received request to `detail-discipline` and any other page that is meant only for TAs or superusers to access.

PR #1325, opened 2018-08-09
Protect `DisciplineDetailView`

alextricity25:
The class-based `DisciplineDetailView` in the lifestudies module
has been protected using the GroupRequiredMixin Django feature.

This commit makes it so only Users in the training_assitant group
can access this view.
A "Forbidden" message will appear otherwise if a User who is not
a in the group tries to access the view.

Fixes: #1323


alextricity25:
@iamsitting I updated the commit to use `GroupRequiredMixin` as requested. I'll also spin up another PR to protect some of the other views in the lifestudies module regular users shouldn't have access to(e.g approve-summary).

PR #1326, opened 2018-08-09
Room Reservation import script for Fall 2018

zioninzion:


yipingbenjaminliu:
for future scenarios, don't leave the raw csv file containing information sit openly on github. The contents is just data and it shouldn't sit with the codebase unless it's a test set for testing. The script is fine but the csv file should've just been uploaded onto the live server via scp instead of letting it sit here on open source

PR #1336, opened 2018-08-10
Datatables nodrf

bdou:


iamsitting:
I'm merging, but let's keep #1025, so we could eventually move to DRF Datatables.

PR #1363, opened 2018-08-18
update domain for ABSENT_TRAINEE_ROSTER_EMAIL, remove attendanceproj …

davidhtien:
…email from email list, remove debug print statements


ap/ap/settings/prod.py
```diff
@@ -17,7 +17,7 @@
 EMAIL_USE_TLS = False
 DEFAULT_FROM_EMAIL = 'AP <ap@ftta.com>'
```
iamsitting: Let's change these to .org as well

PR #1383, opened 2018-08-27
no future finalization

iamsitting:


yipingbenjaminliu:
not sure if this PR is ready, but from what I'm aware of it may be better to create a PR for the following:
all the below only applies to trainees on self-attendance
- can submit roll for any week which is not finalized
- can only finalize one week once
- can only finalize weeks which have completely passed
- force to personal page for rolls finalization by Wednesday of the following week (exception with end of term dates)

yipingbenjaminliu:
failed test case:
- tried to go to next week and submit absence for a whole day, would not allow me
- was able to finalize the current week even though it's not over

iamsitting:
@yipingbenjaminliu could you review the roll table again. Personal attendance was taken care of in #1424.

davidhtien:
This code works but it seems like finalization doesn't do anything? I pressed finalize and I can still edit the rolls after the page reloads

yipingbenjaminliu:
@iamsitting , @davidhtien 
I seem to be seeing something different, the only case where the finalization for roll table entry is blocked off is for house and team. This is what I get when I change status to one of the HC or TM, submit rolls, then click on finalize. 
image

PR #1384, opened 2018-08-28
new roll create logic

iamsitting:
- Allows AMs to input self-attendance (non-audit) rolls on behalf of the trainee through personal attendance record. (AKA submitted_by == trainee, not the AM)
- More robust roll creation logic

### DIFF
--- a/ap/attendance/serializers.py
+++ b/ap/attendance/serializers.py
@@ -21,9 +21,9 @@ class RollSerializer(BulkSerializerMixin, ModelSerializer):
     list_serializer_class = BulkListSerializer
     fields = ['id', 'event', 'trainee', 'status', 'finalized', 'notes', 'last_modified', 'submitted_by', 'date']
     validators = [UniqueTogetherValidator(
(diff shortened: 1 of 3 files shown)


yipingbenjaminliu:
when submitting rolls using personal attendance, the submitted_by field is generated on the backend and not on the frontend, that's why the serializer used adjusted_data and not data. You should be able to make a small change to adjusted data to allow AMs to input self_attendance rolls on behalf of the trainees

iamsitting:
I'm addressing two independent issues.

First: Self-attendance roll input for AMs. This is addressed by L14 on RollPane.js and L77 on serializer.py. Either (backend or frontend) solution is fine I think. But I wanted to isolate the change so it would only affect the personal attendance record (as oppose to the roll table).

Second: The rest of the changes are to address several ongoing problems. Trainees are able to delete audit rolls if they submit present rolls for the corresponding event (vice-versa for AMs). In some cases, AMs cannot enter rolls if there is already a self-attendance roll.

PR #1393, opened 2018-09-01
1381/reasons required trainees

kcmin:


kcmin:
I tested it as Jerome and it worked on my end. I think we may be looking at two different bugs, let's touch base later this week.

PR #1394, opened 2018-09-03
better url handler

iamsitting:
Sorts gospel trips by most current and gets the first open one.


iamsitting:
On Saving a gospel trip or entering a new gospel trip getting a blank page http://127.0.0.1:8000/gospel-trips/trip/3

iamsitting:
I couldn't duplicate this error. Can you test again @finben ?

PR #1399, opened 2018-09-15
1252/exam fixes

francestsenn:
- Option to make graded exams available separate from making the exam available (e672bcc, aa9d9da, be92ea0)
- Required number of questions to answer can no longer be greater than question count (2fc24de)
- Can save and preview exam while editing (385a3b5, 6f6eb45)
- UI changes for small problems:
- d919ca7 Button to "Add Section" was moved from the top of the page to the bottom
- 11019cf Wording change for a link to "View Exams"
- 768b17a Button for "Save & Preview"
- a57c85a Dropdown for filtering exam list now only shows exams and is in alphabetical order


iamsitting:
@yipingbenjaminliu take a look.

yipingbenjaminliu:
I'm on it

ap/aputils/templatetags/smart_menu.py
```diff
@@ -83,7 +83,7 @@ def generate_menu(context):
       ],
       trainee_only=[
         SubMenuItem(name="Take Exam", url='exams:list', condition=context['exams_available']),
-        SubMenuItem(name="View Graded Exams", url='exams:taken', condition=context['exams_taken']),
+        SubMenuItem(name="View Exams", url='exams:taken', condition=context['exams_taken']),
```
yipingbenjaminliu: Is it possible to merge both options into a single page rather than two separate pages that display different filters? Also, I think the view and taken are flipped for the URl.

Take Exams
image

View Exams
image

ap/exams/context_processors.py
```diff
@@ -32,7 +31,7 @@ def exams_taken(request):
   for session in sessions:
     if session.exam == None:
       session.delete()
-    elif session.exam.is_open:
+    elif session.exam.is_graded_open:
```
yipingbenjaminliu: What's the rationale for elif rather than just if? I'm thinking that if the session doesn't have an exam, it should be deleted and that if the session exam is graded open, it would be added to the exam_count

ap/exams/static/exams/js/exam-create-save-post.js
```diff
@@ -76,4 +76,45 @@ $(() => {
...
+      }
+    });
+
+    $.ajax({
+      type: 'POST',
+      url: window.location.pathname,
+      data: JSON.stringify(rtn_data),
+      success: function(data, status, jqXHR) {
+        if (data.ok) {
+          window.location = PREVIEW_URL || window.location;
```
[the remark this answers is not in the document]
yipingbenjaminliu: there maybe someting wrong with the ajax success as the exam data is saved by doesn't open a preview

ap/exams/context_processors.py
francestsenn: With if instead of elif, it looks like the session without an exam would be deleted and then checked for is_graded_open. Using is_graded_open on a non-existent session.exam would cause an error

ap/aputils/templatetags/smart_menu.py
francestsenn: With only one page, do you think we should put it back in the "Current" tab? It looks dumb to have one thing in the dropdown for "Exams", and having it go directly to Exams when clicking it is inconsistent with all the other tabs

yipingbenjaminliu: Yes.

ap/exams/context_processors.py
yipingbenjaminliu: got it, god point.

yipingbenjaminliu:
@francestsenn , from what I understand, if you click on Create, the object is created and you're redirected to the UpdateView of the same exam. If you click on Create and Preview it would save the object, then it either opens a new window and renders what the trainee would see or redirects to what the trainee would see.

francestsenn:
@yipingbenjaminliu I removed the button for "Create & Preview". Clicking "Create" will bring you back to the exam list page. I'll work on the "Create & Preview" option, but I wanted the module to work well enough for the start of midterms.

davidhtien:
Small issue, when viewing graded exams I get a double

img_0728


Everything else seems good though!

francestsenn:
@davidhtien Last commit should fix the problem

davidhtien:
Waiting for travis to pass but other than that LGTM

ap/exams/forms.py
```diff
@@ -2,22 +2,20 @@
...
 from accounts.widgets import TraineeSelect2MultipleInput
 from datetime import datetime
 
 
 class ExamCreateForm(ModelForm):
   training_class = ModelChoiceField(
-      Event.objects.filter(start=datetime.strptime('10:15', '%H:%M'), type='C')
-      .exclude(name="Session II") | Event.objects.filter(start=datetime.strptime('08:25', '%H:%M')).exclude(name="Session I").exclude(name="Study Roll").exclude(name="Study").exclude(name="End Study"),
-      empty_label=None
+    Class.regularclasses.all(), empty_label=None, label="Training Class"
```
davidhtien: @francestsenn I thought we weren't going to use this manager for now? Did you get it to work somehow?

PR #1406, opened 2018-09-17
Sister ta actions fixes

davidhtien:
- Approve & transfer no longer defaults no matter what to the trainee sister’s TA brother but instead checks the value of the form field

- Change button positions to make usage more intuitive, transfer and approve & transfer along with the transfer to form field and the save, mark for fellowship, and delete buttons above them under the rest of the form fields

- Mark for fellowship should not save the value in the transfer to form field

- Add button for unmarking for fellowship

- Clicking save should save the leave slip and then redirect to the next leave slip url


davidhtien:
Please test everything both in the detail view and the list view for ta sisters and brothers, every button should work according to spec

PR #1407, opened 2018-09-18
Roll Model Manager fix

yipingbenjaminliu:
to set the standard query parameter on rolls to be the current term by using custom manager


iamsitting:
#1352

PR #1414, opened 2018-09-24
making ajax requests synchronous

iamsitting:


davidhtien:
I'm not sure if you're aware of this but you can set the setting "async" to false in the ajax call to make it synchronous. What your doing seems a little excessive? I can't really follow your thought process.

https://stackoverflow.com/questions/133310/how-can-i-get-jquery-to-perform-a-synchronous-rather-than-asynchronous-ajax-re

PR #1418, opened 2018-09-29
1381/rr doublebookfix

kcmin:
When Jerome approves a room reservation request for a room/time that has a conflict, there is no prompt indicating that there was a conflict, giving the impression that the request went through.

Please test it out to see if you guys can double book and whatnot, or if you guys try to put in something valid and it ends up failing.


yipingbenjaminliu:
@kcmin, the clean method wasn't working for my test cases. I changed the logic a little and commented out the old code. Review it and test it and go from there.

yipingbenjaminliu:
@kcmin, I tried submitting a RR that overlapped an existing reservation.
the following two cases were not working
image
image

yipingbenjaminliu:
@kcmin, which test cases failed the current logic?

kcmin:
@yipingbenjaminliu Benji, for both of those test cases, that is the expected outcome. 


Also from your logic, this failed for me:
image

iamsitting:
Tested. Looks good. Let's return better feedback? Also doesn't account for Pending room reservations.

iamsitting:
My bad didn't mean to delete gitignore. Will put back then merge.

Issue #1233, opened 2018-06-08
Absent Trainee Roster reminder


billyeh:
@finben Johnny says you all used to print out a reminder sheet if houses forgot to put attendance in but he can't find that on the server.

finben:
Yeah, we don't have it on the server. I don't know if it needs to be on the
server. If anything it's just a one page word document to be uploaded.

Bill Yeh <notifications@github.com> schrieb am Fr., 8. Juni 2018, 10:20:

yipingbenjaminliu:
or should we notify it as a bell saying that your house hasn't submitted it today?

finben:
@yipingbenjaminliu I don't think the bell would help much. Because they have to submit by 8:00 AM. And anyone logging on the attendance server before 8:00 AM is probably already going to be logging on to submit their house attendance. At least that's my thought.

yipingbenjaminliu:
oh, i mean let the bell sit there for the entire day reminding them that they didn't submit it that day

finben:
Oh I see yeah that does sound good. It would need to sit there for both HC's. Sounds good to me

yipingbenjaminliu:
@bdou, you should probably raise an issue for that, thanks.

JoannaRumbley:
Here's an update for this issue. @poohbear0628 
HCs are still getting notifications about not doing the absent trainee roster on Mondays. But the HCs do not need to fill out this page on Mondays, so it should not be notifying them. As long as the HCs have done the other days, they should not receive a notification.

JoannaRumbley:
PR #1607

JoannaRumbley:
@poohbear0628 
After fellowshipping, we decided to make the notification as a pop up announcement instead of using the notification bell. After 8:25 AM, if an HC forgot to submit it, there will be a pop-up message. It will remain as a pop up on their homepage until they click the Dismiss or Mark as Read button. (It will **not** automatically disappear at the end of the day.) 
(This means that if an HC doesn't log on to the server for two whole days in a row, they will have two of the same notification on the third morning before 8:25 AM. If they still forget and don't even log onto the server the third morning, they'll have a third pop up that they'll have to dismiss.)

Also, @ryanli633 – we need a character limit for each entry’s comment box

PR #1419, opened 2018-09-30
1233/initial annoying reminder

yipingbenjaminliu:
to address #1233


finben:
This looks great. I tested and this does exactly what I think we want. Good job thanks Benji!

PR #1431, opened 2018-10-06
To allow unmarking marked for fellowship for web access

finben:
- [x] Have option to unmark items for fellowship on web access module.


finben:
Thanks Benji, how about now?

PR #1441, opened 2018-10-08
Ticker vertical position fix

zioninzion:
👍


iamsitting:
Looks better 👍 Is it suppose to not scroll anymore?

zioninzion:
If the announcement is short enough, it will not scroll.

PR #1442, opened 2018-10-08
Designated services force page

AngeloBelUCSD:


yipingbenjaminliu:
before this gets merged in, we need to address the numerous IndexError for certain cases as seen in the gmail account error logs. The last one was from Sep. 29. There needs to be a mechanism to reflect how the error is handled rather than just a generic error page. I think the error has something to do with the services record being absent, not sure about the details but someone needs to look into it and address it before this gets merged in.

yipingbenjaminliu:
@AngeloBelUCSD, actually, the error that the server has been throwing is because people weren't being properly added into designated services. It's technically an independent issue from force page, but since we're at it, we may as well clean up the mechanism a little.

The IndexError thrown to the server was because access to the 'Designated Service Hours' option under 'Personal Attendance' was determined by the group 'designated_service' which was completely separated from everything else. So I simply removed that group and got the access to check for designated service and if there isn't one, the remove access.

ap/services/views.py
```diff
@@ -364,15 +363,15 @@ def get_object(self, queryset=None):
...
 
     # get service
-    try:
+    if 'sevice_id' in self.kwargs.keys():
       self.service_id = self.kwargs['service_id']
-    except KeyError:
+    else:
       self.service_id = self.designated_assignments[0].service.id
 
     self.service = Service.objects.get(id=self.service_id)
```
yipingbenjaminliu: @AngeloBelUCSD, actually, the error that the server has been throwing is because people weren't being properly added into designated services. It's technically an independent issue from force page, but since we're at it, we may as well clean up the mechanism a little.

The IndexError thrown to the server was because access to the 'Designated Service Hours' option under 'Personal Attendance' was determined by the group 'designated_service' which was completely separated from everything else. So I simply removed that group and got the access to check for designated service and if there isn't one, then remove access.

So the code I pushed should resolve all the IndexError caused by this line

iamsitting:
Merged into #1510

PR #1450, opened 2018-10-11
Groupslip fixes and code refactor

yipingbenjaminliu:
@finben, this should address Hannah's concern in the email about saving GroupSlip that removes the TA


davidhtien:
@iamsitting @finben can you guys review and test again and merge if everything seems good?

PR #1475, opened 2018-10-27
change roll serializer to single roll system, add management command

davidhtien:
Needs to be tested


davidhtien:
@yipingbenjaminliu took a quick glance, seems that code still supports audit rolls? Does it account for changing people onto self-attendance halfway through the term?

I made this PR because we talked to the TAs and they feel good that we switch to keeping audit rolls on paper, so only roll for each event at any given time

yipingbenjaminliu:
ah, okay. I'm out of the loop then, please disregard that PR then.

ap/attendance/management/commands/change_to_single_roll_system.py
```diff
@@ -0,0 +1,39 @@
...
+      roll = rolls[i]
+      trainee = roll.trainee
+      event = roll.event
+      date = roll.date
+
+      duplicates = Roll.objects.filter(trainee=trainee, event=event.id, date=date)
+      if duplicates.count() == 1:
+        dup = duplicates.first()
+        if dup.submitted_by != trainee:
+          dup.submitted_by = trainee
```
iamsitting: I think we should preserve the submitted_by. Is there a reason not to?

ap/attendance/management/commands/change_to_single_roll_system.py
```diff
@@ -0,0 +1,39 @@
...
+          dup.save()
+      else:
+        for dup in duplicates:
+          if dup.submitted_by != dup.trainee and dup.submitted_by in AMS:
```
iamsitting: What about just (this will take of rolls inputted by team monitors and HCs):
```suggestion
          if dup.submitted_by != dup.trainee:
```

ap/attendance/management/commands/change_to_single_roll_system.py
davidhtien: We can't because of the constraint 😢

ap/attendance/management/commands/change_to_single_roll_system.py
davidhtien: According to Sven HCs and team monitors can't submit audit rolls so we don't need to worry about that

yipingbenjaminliu:
@iamsitting , the above two commits should resolve those problems

PR #1480, opened 2018-10-31
Classnotes fix

samvictor232:
Service leaveslips were not being found because the end time for the leaveslips was before the end time of the event. 
For example:
Breakfast clean up service slip:
Start: 7:50am
End: 9:30am
CYPC:
Start: 8:25am
End: 9:59am
Changed to: If the group slip covers the start time of the event, then the roll is excused. 
^Not sure if that's the preferred solution.

### DIFF
--- a/ap/classnotes/utils.py
+++ b/ap/classnotes/utils.py
@@ -97,7 +97,7 @@ def get_leaveslip(trainee, roll):
   individualslips = IndividualSlip.objects.filter(trainee=trainee, status='A', rolls__in=[roll])
   roll_start_datetime = datetime.combine(roll.date, roll.event.start)
   roll_end_datetime = datetime.combine(roll.date, roll.event.end)
-  groupslips = GroupSlip.objects.filter(trainees=trainee, status='A', start__lte=roll_start_datetime, end__gte=roll_start_datetime)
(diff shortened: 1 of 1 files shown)


ap/classnotes/utils.py
```diff
@@ -97,7 +97,7 @@ def get_leaveslip(trainee, roll):
   individualslips = IndividualSlip.objects.filter(trainee=trainee, status='A', rolls__in=[roll])
   roll_start_datetime = datetime.combine(roll.date, roll.event.start)
   roll_end_datetime = datetime.combine(roll.date, roll.event.end)
-  groupslips = GroupSlip.objects.filter(trainee=trainee, status='A', start__lte=roll_start_datetime, end__gte=roll_end_datetime)
+  groupslips = GroupSlip.objects.filter(trainees=trainee, status='A', start__lte=roll_start_datetime, end__gte=roll_start_datetime)
```
iamsitting: I don't know if it catches every case. How about...
```suggestion
  groupslips = GroupSlip.objects.filter(trainees=trainee, status='A', start__lte=roll_end_datetime, end__gte=roll_start_datetime)
```
See https://github.com/attendanceproject/djattendance/blob/99884f39a1486515692d4cb3d1d8367564b428af/ap/accounts/models.py#L388

samvictor232: Yeah that looks good

PR #1510, opened 2018-11-21
Designated Service Force Page

iamsitting:
- [x] Add Workers to Services
- [x] Generate Service Hours for previous weeks
- [x] Force page to designated service hours


AngeloBelUCSD:
Tested addition of trainees to designated service, excused status, and management command.
All work.

PR #1518, opened 2018-11-30
1252/exam fixes

francestsenn:
Addresses two issues found with grading:
- "When trying to manually enter scores for handwritten exams via "Grade for paper response," after we enter a number and click "Save Grades," we also get directed to a "504 Gateway timeout" error page for some trainees." Errors look like they're caused because a session already exists for the trainee and the exam. For example, a trainee may have saved but not finalized their exam. Therefore, when entering a paper response grade, the new session could not have been created.
- "When grading exams, we will get "504 Gateway timeout" or "500 Lost Sheep" errors for some trainees when we try to open their responses via "Grade Exam," and we are not able to view their midterms." The only problem I could duplicate was in opening Johanna Findeisen's New Jerusalem midterm. For Johanna's test, the issue was that the graders' comments included quotation marks, which caused errors.


ap/exams/utils.py
```diff
@@ -400,7 +400,8 @@ def save_grader_scores_and_comments(session, section, responses):
   if section.section_type == 'E' and responses['comments'] == "NOT GRADED YET":
     responses_obj.comments = "GRADED"
   else:
-    responses_obj.comments = responses['comments']
+    comments = responses['comments']
```
yipingbenjaminliu: take a look at the the utf encoding either on rendering the page or inputting the comments to resolve it. More issues may come up if you just try and replace the problematic characters.

ap/exams/views.py
```diff
@@ -249,14 +249,22 @@ def post(self, request, *args, **kwargs):
...
         # Save grades for trainees who use paper submission
-        session = Session(
-            exam=exam,
-            trainee=trainee,
-            is_submitted_online=False,
-            time_finalized=datetime.now(),
-            is_graded=True,
-            grade=float(grades[index]))
-        session.save()
+        try: # Session needs to be unique. If the session already exists, it will modify the existing session.
```
yipingbenjaminliu: use update_or_create instead of try-except. take a look at #1448 and #1448, you can close those two PRs when you're done

Issue #1525, opened 2018-12-01
TA Issues - 12/1/2018


JoannaRumbley:
**Attendance Rolls**
- [ ] Add force page when submitting attendance late. 
- [ ] Button to change attendance period is still only one-pixel large on Firefox. It isn't intuitive either.

**Class Notes**

**Extra Fields**
- [x] Does not affect perfect attendance will be a built-in leave slip category” (Automatic for Conferences, team trips, house meals, service, fellowship, and allow for TA override**) @joysyu 

**Gospel Trips**
- [ ]  Trainees report getting a blank page after trying to submit (specifically we believe this is tied to Chrome browser)
- [ ] When generating a report, if I generate a report containing "Allergies" plus any other field, the ability to search and generated CSVs, etc. disappears.
- [ ] Add functionality for us to add finance coordinators, media coordinators, and report/statistics coordinators. It would be in the same principle as the trainee contacts.
- [ ] I have just inputted the trainee contacts for the gospel trips. When looking at the report on the webpage, a check mark shows. However, when downloading the CSV, nothing shows. Can there be a simple “Yes” indicator? Along these lines, “team contact” should be renamed “trainee contact”.

**Leave Slips** 
- [x] Automatically approve when the TA submits as leaveslip. @davidhtien #1534 
- [x] Have leave slips show the date and time they were submitted.
- [ ] **Critical** Issue of multiple leave slips for the same roll call cancelling one another. For example, I approve a leave slip for a roll call, but the attendance does not reflect the approval because there is another leave slip for the same roll call that was not processed or denied. 
- [ ] When a leave slip is marked for fellowship have a note of who marked the leave slip for fellowship, so the trainee knows who to go to. (We need to visit this along with the bigger idea of a messaging module)
- [ ] There are still issues with how the attendance server processes leave slips for the same roll call(s). Example: Celine Goh submitted four leave slips for Lord's Day curfew and/or lights out for 9/23. One was approved. However, when you pull up any of the other three leave slips, the roll call still shows the absences.
- [ ] When pulling up a leave slip, the calendar only shows the absences and tardies for that leave slip. We would like to see the trainees' complete attendance. In other words, when viewing the weekly calendar, the trainee attendance view should be the same as the leave slip view.
- [ ] Add "Does not count against perfect attendance" check box and functionality.

      
**Life Studies**
[...]

PR #1526, opened 2018-12-02
Leaveslips that don't count against perfect attendance

joysyu:
-added a 'does_not_count' field to the leaveslip model
-when a leaveslip is created w/ category conference, fellowship, team trip, or service, the leaveslip automatically does not count against perfect attendance
-a TA can use a checkbox to toggle whether the leaveslip counts against perfect attendance
-includes script to update existing leaveslips


joysyu:
addresses #1525

ap/leaveslips/views.py
```diff
@@ -80,6 +80,7 @@ def post(self, request, **kwargs):
       update['ta_sister_approved'] = True if request.POST.get('ta_sister_approved') == 'on' else False
     if request.POST.get('texted'):
       update['texted'] = True if request.POST.get('texted') == 'on' else False
+    update['does_not_count'] = True if request.POST.get('does_not_count') == 'on' else False
```
bdou: can we wrap this statement in if statements like the other ta_sister_approved and texted or do you think it doesn't matter?

libraries/react/scripts/actions.js
```diff
@@ -314,6 +314,7 @@ export const postLeaveSlip = (values) => {
     "host_name": values.hostName,
     "host_phone": values.hostPhone,
     "hc_notified": values.hcNotified,
+    "does_not_count": (values.slipType.id == 'CONF' || values.slipType.id == 'FWSHP' || values.slipType.id == 'SERV' || values.slipType.id == 'TTRIP') ? true: false,
```
bdou: I think it can just be assigned to (values.slipType.id == ...) without the ternary operator? ditto below.

joysyu:
Current tag name is "Does not count against perfect attendance". Since it's really long, we're thinking of other potential names: "No penalty", "No discipline", "Not counted", "Perfect Attendance". Would you be able to ask the TAs please LOL thank u :) @finben

libraries/react/scripts/actions.js
joysyu: Thanks!!

joysyu:
@bdou I agree that we should try to merge this in before next term. Attendance report filters by `does_not_count` instead of leaveslip type, so that wouldn't work with the existing leaveslips on the server.

ksungreborn:
Not sure if it's just with my machine but I try and create a leaveslip on the trainee side, and I get a 403 error, and am unable to create a leaveslip?

ap/leaveslips/management/commands/populate_perfect_attendance_leaveslips.py
```diff
@@ -0,0 +1,13 @@
+from django.core.management.base import BaseCommand
+from leaveslips.models import IndividualSlip, GroupSlip
+
+class Command(BaseCommand):
+  # to use: python ap/manage.py populate_perfect_attendance_leaveslips --settings=ap.settings.dev
+  def _update_leaveslips(self):
+    LS_TYPES = ['SERV', 'CONF', 'TTRIP', 'FWSHP']
+    individualSlips = IndividualSlip.objects.filter(type__in=LS_TYPES).update(does_not_count=True)
+    groupSlips = GroupSlip.objects.filter(type__in=LS_TYPES).update(does_not_count=True)
+
```
iamsitting: This could be helpful: https://docs.djangoproject.com/en/2.1/ref/models/querysets/#update

joysyu: fixed? this is what you're talking about, right?

iamsitting: 👍

iamsitting:
@ksungreborn That may have been a csrf token issue, it was taken care of in dev branch. So it should be good now.

PR #1528, opened 2018-12-04
Gt new fields

iamsitting:
- This will drop the team contact field, so merge this next term.


kcmin:
Tested and it seems to work with the new changes

PR #1532, opened 2018-12-08
fix name input requirement

iamsitting:


iamsitting:
Form_access works, but trainee select form doesn't.

PR #1533, opened 2018-12-08
show all trainees with gt app

iamsitting:


finben:
Works like a charm! LGTM

Issue #1546, opened 2018-12-19
Self-attendance not finalizing


JoannaRumbley:
Ady Garcia has not been able to finalize her attendance for 2-3 weeks. After she clicks Finalize, the notification still shows up on the home page. She uses Chrome, so it shouldn't be a browser issue.
She talked with Megan (attendance monitor), who reproduced the same issue on Ady's account. Megan was able to help her finalize, but she's had to help her with the same issue for 2-3 weeks now. Apparently, Megan said this has happened to a few trainees.

joysyu:
Doesn't seem to be a problem for now.

PR #1558, opened 2018-12-21
preventing duplicates in rolls finalization

joysyu:
prevents duplicates in the string of finalized weeks


joysyu:
related to #1546

PR #1568, opened 2018-12-22
Homepage service row

berlyseria:
added the service row on the front page. On hover, it will show what service you have that day.


appletreeisyellow:
@mycodemin Request for code review

PR #1569, opened 2018-12-22
fix css for designated service hours- not finished yet

soaustin:
This will add a blue banner across the site. Which is something that is on every page of the mockups, but has not been implemented yet. 
I also started editing the designated service hours page, but it is not finished.


JoannaRumbley:
Can something be done about the drop down when you click the textbox? The drop down (the one that automatically suggests date-times based on what you inputted before) is completely blocking the calendar. This makes it very very tedious to input dates/times. @kcmin or @whoever wants to take this on :)

PR #1581, opened 2019-02-15
Gospel stats refactored

iamsitting:
@zioninzion I've refactored your code to take advantage of Django's QuerySet API and to optimize our database queries. The logic should still function the same.
Merging this won't merge to dev, it will merge to your gospel stats branch.
BTW, nice work :+1:


zioninzion:
It seems to be a lot slower when I select multiple teams and multiple weeks to generate report

iamsitting:
@zioninzion hmmm... Ok, I'll look at it again.

PR #1584, opened 2019-02-16
Gospel stats refactored

iamsitting:
@zioninzion I performed some timeits, it looks like the calculations has some initial overhead, but this will work better as the reports deal with more data.

However, generating the pdf files takes a long time. But that's the case before and after I made changes.


Plew:
Zion has stated that we will no longer be refactoring gospel stats, closing this PR.

Issue #1599, opened 2019-03-03
Roll Entry Seating Chart not showing main classes


joysyu:
Attendance monitors should be able to select roll for main classes through the roll entry seating chart.

joysyu:
Christine deleted the main class seating chart and then re-created it. When she re-created it, none of the events' chart properties were updated. Now, none of the main class events are linked to the new seating chart.

Plew:
Closed in #1606

PR #1606, opened 2019-03-09
link main class events to seating chart

joysyu:
Currently, main class events aren't linked to any seating chart. The issue of linking new seating charts to events still needs to be fixed; this is a quick fix for now so attendance monitors can save a lot of time. #1599


joysyu:
@ryanli633 When new charts are created, I don't think any events are updated to link to that chart; this causes issues like #1599. What do you think would be the best way to fix this? Should the Chart model have a class type field?

PR #1607, opened 2019-03-12
update absent trainee roster

poohbear0628:
notifications about not doing the absent trainee roster on Sundays and Mondays won't show up again after the update


bdou:
Hi Serena! One note: HCs are still required to put in house attendance on Sundays.

poohbear0628:
@bdou okay! just made the change to the code. Has will still receive the notifications on Sundays butn not on Mondays.

JoannaRumbley:
When will the change be made on the live server? It still showed the notification today for Monday

ryanli633:
I deployed the change on the live server on Saturday. So, it should have worked for this Monday. I didn't really look through the code in detail though. @poohbear0628 , could you double check to make sure the alert doesn't show on Mondays?

PR #1612, opened 2019-03-16
Fixed web request deleting

Plew:
YES


kcmin:
Checked to see that this works correctly both from trainee / TA views

PR #1623, opened 2019-04-03
Leaveslips: redirect TAs to TA view

joysyu:
When a TA clicks on a leaveslip in the summary pane, the TA detailed view of the leaveslip will open in a new window.


ryanli633:
So, I merged this in without testing it, but it appears to me that it does not work on the live server? I tried a couple different leave slips under Jerome, and I still saw the trainee view instead of the TA view.

PR #1624, opened 2019-04-04
Greek helper

berlyseria:


ap/greek/templates/greek_helper/_flashcards.html
```diff
@@ -0,0 +1,144 @@
...
+  let greekVocabList_fc = [];
+  let startingChapter_fc = 1;
+  let endingChapter_fc = 1;
+  let currentVocabIdx_fc = 0;
+
+  function updateIndexText() {
+    $("#flashcard-index").text(`${currentVocabIdx_fc+1}/${greekVocabList_fc.length}`);
+  }
+
+  function updateArchiveCheckbox() {
```
mycodemin: Not a big deal, but for that minimal performance boost you could have the function default to show, then do a check to see if you should hide.

this could apply to some of the other functions as well. I know it reads better doing it this way, but I'll leave it up to your discretion

PR #1628, opened 2019-04-06
General improvements to room reservations page

kcmin:
1. Removed (hid) the duplicate date under approved room reservations. 
2. Added some placeholders for the room reservations page.
3. Added some JS and struggled deeply on the the start and end times but I am going to figure out how to fix it later or let someone else fix it. But it should automatically round to nearest half-hour.


kcmin:
I know codacy is failing but is it necessary to fix these issues? maybe it'd be good to leave those in.

PR #1630, opened 2019-04-06
Interim intentions

bdou:
Modified interim intentions page to match design on dropbox; fixed interim itinerary logic and functionality.


francestsenn:
Logic is good (thumbs up)
Text box for notes/comments is short, which only seems to be a problem on Chrome

Plew:
I am merging now because this is being used currently, but if someone could fix this text box issue that would be awesome! I dont have much HTML experience.

PR #1632, opened 2019-04-06
Check scheduling exceptions

joysyu:
Add page to displays a table of service assignments that violate active exceptions.


joysyu:
@ryanli633 What specifically should be on the page? Right now, each row of the table is service name and the names of trainees. Also, do we want to be able to directly edit service assignments from this page?

kcmin:
It works as it should for all active exceptions. It runs a little slowly but that's just the nature of the check. I talked with Joy about making the exceptions more descriptive

PR #1633, opened 2019-04-06
Added Term field to Rolls, so that previous term's rolls don't come up.

ik1122:
attendance/views.py 		added "term = CURRENT_TERM" to all instances of RollsFinalization
attendance/notifications.py added "term = CURRENT_TERM" to all instances of RollsFinalization
attendance/models.py		added term = models.ForeignKey(Term) to RollsFinalization class


Plew:
Fixes problem with Daniel Lim

PR #1634, opened 2019-04-07
1252/exam fixes 2

francestsenn:
- When exam was unfinished before the testing period was closed and a makeup was opened, the trainees' saved responses should appear.
- See if exams are opened/closed from the exam list page.
- Open/close exams from the exam list page.
- Allows " character in the graded exam's comments.


AngeloBelUCSD:
All the above works. But, if make up exam is open, and we open the graded exams, there will be two exams shown.

francestsenn:
Modifications since you last reviewed the code @AngeloBelUCSD 
- Exam only shows once if both the graded exam and the makeup is open
- Trainees can view their answers after finalizing the exam (but cannot save/re-finalize)
- TAs previewing the exam will see the exam as the trainees do. We previously had two files that were almost identical for this purpose -- exam_preview.html and exam.html.

francestsenn:
More changes
- Leaving the score for a question blank was causing errors when graders saved their part
- Trainees can view their exam after finalizing

PR #1641, opened 2019-04-27
Only checks the changed weeks or trainees to validate rolls

francestsenn:
Should address the problems when there are many trainees or many weeks in a schedule that is being modified. Hopefully, there aren't any more timeout errors. 😬


kcmin:
Frances I nuked my DB, thenI removed myself on Character. Here are on my errors.

image

PR #1645, opened 2019-04-28
new hc rec form handling

iamsitting:
Simple hc rec form handling logic. Just need some front end validation to avoid duplicate entries, but it should work now. @ryanli633


iamsitting:
Just be aware that there is no validation so it's possible for HCs to submit bad information. In other words, it works smoothly if HCs are competent. 😂

Issue #1646, opened 2019-04-29
HC Recommendation Bug


zioninzion:
@ryanli633 @bdou @francestsenn @JoannaRumbley @Plew 

The issue is that it retains the recommendation objects from previous terms.

Every time you 'submit' two things happen:
- It reorders the list.
- It only shows the first X amount of the recommendations (according to the number of current second/third term trainees in your house).
That is why it creates the current effect.

**Quick Fix**:
- Delete all HC recommendation objects of previous terms (and ask the HCs to recheck their submittal) so that the new recommendation objects the HCs create/update will have the current term of HCRecommendationAdmin.

**Recommended Fix**:
- Only show the HC recommendation objects of current Term (survey_admin.term == Term.current_term())
- Prevent reordering of the objects when submitting the from
- Fix not being able to 'ADD' when having not created any recommendation object yet

zioninzion:
#1645

PR #1647, opened 2019-04-29
current-term-rec

zioninzion:
#1646 
It only shows the current term objects. Still needs to fix unintentional reordering and 'ADD' button.

Should still be functionally complete


zioninzion:
Should be working now

bdou:
Hi @zioninzion your code looks good, v. straightforward; but I added a small Javascript fix I noticed. :)

PR #1658, opened 2019-05-17
Bible reading tracker error

francestsenn:
If the week does not exist yet, the default Bible reading tracker page is for the current week.

Trainee has been getting to the server using ap.ftta.lan/bible_tracker/?week=19, which has been erroring (see the Gmail for the issues)


mycodemin:
The logic looks like it works.
But the "if not week" seems redundant since it'll always return false unless it's week 0.

PR #1666, opened 2019-05-25
Some designated hours and room reservations

mycodemin:
- [ ] Putting an end date before the start date creates unreliable hours. -- a bit higher priority

There's a check now that makes should disallow this

and refactored some code for room reservations


mycodemin:
#1667

PR #1688, opened 2019-06-22
line-85-linens-permission

slin6:
Added linens permissions so linens doesn't get 403 error when marking linens requests as complete.


zioninzion:
LGTM

PR #1692, opened 2019-06-26
update react packages

iamsitting:


davidhtien:
according to this [post](https://softwareengineering.stackexchange.com/questions/340705/when-should-dependencies-be-updated?newreg=10117c32cca444c1b5c673b40a4c36ef) we should upgrade dependencies when either:

1. It's required
2. There's an advantage to do so
3. Not doing so is disadvantageous

React 16 is not required
React 16 doesn't give us any significant advantages (maybe if we move to server side rendering one day we should do this upgrade since there is a significant speed increase on this end)
Staying with React 15 is not disadvantageous to us

PR #1696, opened 2019-06-26
Newly added trainees for designated services

mycodemin:
- [ ] Trainees newly added to designated services are required to input hours from previous weeks they were not on the service. Is there a way to forego this? If not, then please fix the bug not allowing them to input zero hours.

Used the logic from the script that Carlos wrote and implemented it afterwards adding new trainees in.

### DIFF
--- a/ap/services/forms.py
+++ b/ap/services/forms.py
@@ -13,7 +13,7 @@ from services.models import (Assignment, Category, Qualification,
                              ServiceAttendance, ServiceException, ServiceRoll,
                              Worker, WorkerGroup)
 from terms.models import Term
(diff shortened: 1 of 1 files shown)


mycodemin:
Attendance monitors add via /services/add_trainees_services/
Testing on my end didn't give me any force pages. I think it should be ok, but maybe there is some edge case I'm not considering.

Also, all force pages are not enforced on local. You need to go to ap/settings/dev.py and change 1 line, add 1 line:
+ DEBUG = False
+ ALLOWED_HOSTS = ['*']

finben:
Does not work. Failed testing.

PR #1698, opened 2019-06-27
Exam line breaks and saving/finalizing exams

francestsenn:
Line breaks displayed for essay responses and questions, so multiple paragraphs should show up

If exam is closed, the trainee cannot finalize the exam. Instead, it saves. Exams can only be finalized when the exam is open or if a makeup is open.

### DIFF
--- a/ap/exams/templates/exams/essay_question_template.html
+++ b/ap/exams/templates/exams/essay_question_template.html
@@ -10,7 +10,7 @@
 <b>Word Count: </b><span class="char_count"></span>
 <div class="clearfix"></div>
 {% else %}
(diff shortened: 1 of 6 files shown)


francestsenn:
@ksungreborn Sorry I should've clarified about the exam being closed. The issue was for when the exam was closed while the trainee was already working on the exam. They still could finalize after the exam was closed.

ksungreborn:
Okay, so if I understand correctly:
1. Trainee is taking exam.
2. TA closes exam.
3. Trainee tries to finalize and is unable to, but the exam progress is still saved.

If this is the correct behavior, then it looks good.

PR #1702, opened 2019-06-27
Grad Outline new speaking field

bdou:
added a speaking participation field to outline because it may change.


Plew:
Everything works correctly, but I am a little confused on the purpose of this change. However, I dont know much about the grad module. Is it possible to explain the functionality of this?

bdou:
When it comes to graduation speaking, initially all the fourth-termers fill out what outline point they want to speak on and whether it is more truth-based or experience-based. Then the office comes in and assigns speakers, and they may change any and all of that. As a result, the initial entry is different from what their roles end up being. You can speak to Hannah if you want more clarification about it.

PR #1706, opened 2019-06-28
ensure TA is chosen for a leaveslip approved by TA

ksungreborn:
Makes sure a TA is chosen if in leaveslip the trainee selects "Yes, informed a TA"


finben:
Initially seems to work. However when I then select a TA and hit submit it does not seem to submit but just says there hanging.

ksungreborn:
Yeah I noticed. Trying to figure it out

PR #1707, opened 2019-06-28
Makes the new inputs blank when adding designated services

francestsenn:
Adding hours for designated services does not simply clone the last row


francestsenn:
@mycodemin can you look at it again?

mycodemin:
Looks good with the new changes as well

PR #1752, opened 2019-09-14
TA Sisters unable to transfer group leave slips

mycodemin:
Removed a parameter, "sister-save", that ignores when a sister TA would save something which is how they would be able to choose a person to transfer it to. I think without any prior commenting, it's hard to say why it was there in the first place - it might be a safety precaution but either way I don't foresee it being an issue.

EDIT:
Actually we talked to David about this and he's wondering why it would ever get assigned to a brother leaveslip would ever get assigned to a sister TA. Probably fix that bug. The fact that TA sisters only have the option to accept or deny probably is a side-problem... I'll just add a "transfer" button in the event that it happens.

EDIT again:
I replicated the behavior of a TA sister getting a group leave slip that a brother submitted. When a group leave slip is not processed yet, if a trainee sister on the group leave slip changes anything, then it gets transferred over to the trainee sister's TA sister. That being said, I think since it's not a common problem it's pretty low priority. TA sisters can just transfer it back to a TA brother now.

To test:
Check if TA sisters assigned group leave slips from trainee brothers are able to transfer it another TA.

e.g. leaveslips/group/update/29337


Plew:
Issue with the transfer button in the list view, although the detailed view transfer does work.

mycodemin:
Yeah I got that same error. I'll be closing this until I figure out how to fix it.

PR #1753, opened 2019-09-14
no-div-zero

zioninzion:
If zero, set to 1


zioninzion:
Deals with attendance report problem.

PR #1782, opened 2019-10-03
Fixed logic so that TA brothers can approve or deny.

mycodemin:
TA brothers were saying that they were only able to Transfer leaveslips due to the slip in logic. I fixed the logic and added in some comments to make it more clear. I did test it and it should work properly now.


zioninzion:
LGTM

PR #1806, opened 2019-10-26
Schedule weeks bug fix

mycodemin:
Modified the code that weeks are now able to be checked dynamically. 

Previously, it would just take a date__range of the earliest week and latest week in schedule, regardless if you have weeks excluded between them. There would be problems, if for example, weeks 4,5,6 weren't supposed to be included, but weeks 1 and 17 are. Now it can check multiple date ranges.

Also there was an ambiguous keyerror when trying to update or delete schedules, but I haven't gotten those after putting in the fix.

You can learn more about Q functions:
https://docs.djangoproject.com/en/2.2/topics/db/queries/


francestsenn:
It seems to work and addresses the Key Error issues we were getting, but I think the code for the UpdateScheduleForm needs to be changed because I don't think it does what you intended.

mycodemin:
@francestsenn  I ran into a bug error without modifying the code in UpdateScheduleForm. Do you mean change the code as in remove the new changes? Or are there unintended effects from the new code?

PR #1811, opened 2019-11-02
Schedule changes

francestsenn:
- Doesn't automatically deny leave slips from schedule changes
- Deletes leave slip if it only includes events no longer on the schedule
- Afternoon class schedule changes at /schedules/afternoon_changes


francestsenn:
Merged in dev. Deleted Greek II event from the database, which was causing the issue.

